PR #152, opened 2024-01-24
chore: add helikon bootnodes

lrazovic:
Companion PR: https://github.com/Polimec/polimec-binaries/pull/1

### DIFF
--- a/chain-specs/polimec-raw-chain-spec.json
+++ b/chain-specs/polimec-raw-chain-spec.json
@@ -4,7 +4,9 @@
   "chainType": "Live",
   "bootNodes": [
     "/ip4/141.95.47.7/tcp/30333/p2p/12D3KooWSCjh8tN5yzmLgxWR4kauD5NmCmP2ymc3D5WubvMETg6b",
-    "/ip4/57.128.20.108/tcp/30333/p2p/12D3KooWEe7ps1v1mfhjMNaSaukgxsFVL6cqeEQudqZ5t3oQiXNL"
+    "/ip4/57.128.20.108/tcp/30333/p2p/12D3KooWEe7ps1v1mfhjMNaSaukgxsFVL6cqeEQudqZ5t3oQiXNL",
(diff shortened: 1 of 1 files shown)


vstam1 (commented):
Do we need both the wss and normal helikon boot node? Think we can remove one of them (not sure). If not, then I approve as is.

lrazovic:
The P2P WSS is the only method accessible by light clients (src: https://wiki.polkadot.network/docs/maintain-bootnode#accessing-the-bootnode). The standard P2P port is commonly used by other nodes for syncing the network. So, as I understand it, it seems reasonable to maintain both for optimal functionality.

kukabi:
@lrazovic that's right, P2P WSS is included for light clients, but it can be removed if not useful at the moment.

lrazovic:
Thanks for the feedback! 

> @lrazovic that's right, P2P WSS is included for light clients, but it can be removed if light clients are not targeted at the moment.

But there is no downside to including it as of now, is there?

kukabi:
> Thanks for the feedback!
> 
> > @lrazovic that's right, P2P WSS is included for light clients, but it can be removed if light clients are not targeted at the moment.
> 
> But there is no downside to including it as of now, is there?

I think it's safe to include it, the original PR had two lines of test, the second one is for the P2P WSS endpoint:

```
./polimec-parachain-node --chain ./polimec-raw-chain-spec.json --tmp --reserved-only --reserved-nodes "/dns4/boot.helikon.io/tcp/15324/wss/p2p/12D3KooWT2y3C2nzKGJV1Jss3EyCcwGWeroDFfFdJ5CUVz3iQGJD"
```

So the node is able to get up and running just using the P2P WSS endpoint of the boot node. We also incude the same for all relay chain and system parachain boot nodes, example here: https://github.com/paritytech/polkadot-sdk/blob/a9992dbb311862e3698b59f701bc5c4182acb9d7/cumulus/parachains/chain-specs/asset-hub-kusama.json#L19

vstam1 (approved):

PR #154, opened 2024-01-24
Feature/plmc 398 add treasury pallet to base runtime should like the one in

vstam1:
TODO:
- [x] check try-runtime working

Adds the following pallets to base runtime:
- Treasury
- Democracy
- Collective:
  - Council
  - Technical Committee
- Election Phragmen
- Scheduler
- Preimage

Adds pallets to test runtime:
- Election Phragmen

### Democracy 
- From Reserve/Lock -> Hold/Freeze
- Vote lock maximum from `free_balance` -> `total_balance` (allows voting with staked balance)
### Election phragmen
- From Reserve/Lock -> Hold/Freeze
- Remove Deposit (reserve) requirement for voting
- Vote lock maximum from `free_balance` -> `total_balance` (allows voting with staked balance)
- Introduce vote locking period to lock users once they voted.
  - Config item `VotingLockPeriod`: Time users are locked for from the moment they vote.
  - Change in vote -> renews the locking period
  - Can only remove themselves as voters after `VotingLockPeriod` 

## Other important changes
- Added ToGrowthTreasury struct to both base/test runtime. Implements new `OnUnbalanced` that works together with the fungibles traits, to allow sending slashed funds in both Elections and Democracy pallet to the treasury. 
- Changed fast-gov feature to fast-mode feature.
- Introduced instant-mode for Integration testing.
- Removed try-runtime + benchmark features from integration tests

### Treasury Config:
- ApproveOrigin: Either root or unanimous council vote. So either council can directly approve or can be approved via governance.
- RejectOrigin: Either root or >1/2 (5/9) vote. Council majority can reject treasury proposal.


JuaniRios:
We only fixed the tests to use the new fungible API, but didn't actually write a test where we reserve an amount, and can still vote on that right?

I saw the governance tests on `integration-tests` but didn't see us using a reserve and then voting, or a reserve and then electing

vstam1:
> We only fixed the tests to use the new fungible API, but didn't actually write a test where we reserve an amount, and can still vote on that right?
> 
> I saw the governance tests on `integration-tests` but didn't see us using a reserve and then voting, or a reserve and then electing

Important thing to note it that we only use reserves in specific places (deposit for treasury proposal for example) and maybe use Holds. We know that Holds and locks/freezes can overlap. Also locks and freezes can be set after reserves, only reserve cannot be used on same tokens that are already locked. However, I will try to write some more comprehensive tests for the specific vote cases in combination with holds etc.

pallets/democracy/src/weights.rs
```diff
@@ -0,0 +1,987 @@
...
+	fn remove_vote(r: u32, ) -> Weight;
+	fn remove_other_vote(r: u32, ) -> Weight;
+	fn set_external_metadata() -> Weight;
+	fn clear_external_metadata() -> Weight;
+	fn set_proposal_metadata() -> Weight;
+	fn clear_proposal_metadata() -> Weight;
+	fn set_referendum_metadata() -> Weight;
+	fn clear_referendum_metadata() -> Weight;
+
+	fn v2_migration_base() -> Weight { Weight::zero()}
```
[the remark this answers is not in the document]
vstam1: Was from an old migration that was also benchmarked. We didn't need the migrations so this should be removed. Thanks for spotting

integration-tests/src/tests/build_spec.rs
```diff
@@ -40,7 +40,7 @@ fn build_spec_testing_node() {
 	assert_eq!(
 		output.status.success(),
 		true,
-		"Make sure you compile the polimec-parachain-node with \"--release\" and \"--features std,fast-gov\" before running this test."
+		"Make sure you compile the polimec-parachain-node with \"--release\" and \"--features std,fast-mode\" before running this test."
```
lrazovic, line 43: I get:

> error: none of the selected packages contains these features: fast-mode

Should this be: 

```suggestion
		"Make sure you compile the polimec-parachain-node with \"--release\" and \"--features std,instant-mode\" before running this test."
```
?

P.S: Even with `cargo t --release --features std,instant-mode` the test is failing

runtimes/base/src/custom_migrations.rs
```diff
@@ -19,49 +19,26 @@ use crate::*;
...
-impl frame_support::traits::OnRuntimeUpgrade for CustomOnRuntimeUpgrade {
+pub struct InitializePallet<Pallet: GetStorageVersion<CurrentStorageVersion = StorageVersion> + PalletInfoAccess>(sp_std::marker::PhantomData<Pallet>);
+impl<Pallet: GetStorageVersion<CurrentStorageVersion = StorageVersion> + PalletInfoAccess> frame_support::traits::OnRuntimeUpgrade for InitializePallet<Pallet> {
 	#[cfg(feature = "try-runtime")]
 	fn pre_upgrade() -> Result<Vec<u8>, DispatchError> {
-		use frame_support::pallet_prelude::StorageVersion;
-
-		let membership_pallet_version = StorageVersion::get::<OracleProvidersMembership>();
-		log::info!("OracleProvidersMembership migrating from {:#?}", membership_pallet_version);
+		log::info!("Elections migrating from {:#?}", Pallet::on_chain_storage_version());
```
lrazovic, line 28:
```suggestion
		log::info!("{} migrating from {:#?}", Pallet::name() ,Pallet::on_chain_storage_version());
```

runtimes/base/src/custom_migrations.rs
```diff
@@ -19,49 +19,26 @@ use crate::*;
...
 		Ok(Vec::new())
 	}
 
 	#[cfg(feature = "try-runtime")]
 	fn post_upgrade(_state: Vec<u8>) -> Result<(), DispatchError> {
-		use frame_support::pallet_prelude::StorageVersion;
-
-		let membership_pallet_version = StorageVersion::get::<OracleProvidersMembership>();
-		log::info!("OracleProvidersMembership migrated to {:#?}", membership_pallet_version);
+		log::info!("Elections migrated to {:#?}", Pallet::on_chain_storage_version());
```
lrazovic, line 34:
```suggestion
		log::info!("{} migrated to {:#?}", Pallet::name(), Pallet::on_chain_storage_version());
```

lrazovic (changes requested):
Few minor changes needed, but overall looks fine. As soon we are aligned on the runtime config, we can merge this.

pallets/democracy/README.md
```diff
@@ -0,0 +1,141 @@
+# Democracy Pallet
+This repository contains a modified version of Parity's 'democracy' Substrate Pallet. The original version can be found [here](https://github.com/paritytech/polkadot-sdk/tree/master/substrate/frame/democracy).
```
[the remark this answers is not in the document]
vstam1: Not sure if that is useful for democracy as it might still be an actively maintained branch (used PJs branch for moving to fungibles as base).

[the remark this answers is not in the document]
lrazovic: Exactly, because it is still maintained, it would be more sensible to link to the specific commit/branch/tag we used as a base, rather than the master, which could change and render this README outdated.

integration-tests/src/tests/build_spec.rs
vstam1: Hmmm which command are you actually running? We added the default-members so everything that does not have a clear specified -p will only use `nodes/*` and `pallets/*` so this test shouldn't be running if we call cargo test. Tbh I do not think this test should be in the integrations test folder as it requires the node build with specific features, that are now also different from the default integration test features. Might be better to move it to some new package like "UI testing" that is specifically build to do Multi project instantiations for front-end testing. 

But I can try to fix the test and feature flags. Wdyt @JuaniRios ?

JuaniRios: We can build the node with instant or fast mode, it doesn't matter.
But it's a valid point that this test might not belong there if it requires some previous step in order to pass.

Leo's problem was that he forgot to compile the node before running that test.

We discussed also maybe adding a #[ignore], and then running it manually if we ever want to, although would be nice to have it automated so we never forget. Open to any other suggestions

vstam1: I think it would be nice to create a separate folder for create multi project nodes etc where we have a specific Readme + this test + maybe all the necessary cargo requirements.

vstam1: For now we can do an #[ignore]

lrazovic: `cargo b -p polimec-parachain-node --release --features std,fast-mode && cd integration-tests && cargo t build_spec_testing_node -- --nocapture`

> Members length cannot exceed MaxMembers.

image

I think it's because we set 6 members in the technical council while the maximum is 5.

vstam1: Interesting. Thanks I'll take a look!

vstam1: Okay fixed. Was indeed max 5 technical committee problem.

lrazovic: Tested with `cargo t build_spec_testing_node -- --include-ignored`. It works!

lrazovic (approved):

integration-tests/src/tests/governance.rs
```diff
@@ -0,0 +1,475 @@
...
+	PolimecBase::execute_with(|| {
+		let alice = PolimecBase::account_id_of(ALICE);
+
+		assert_ok!(Balances::reserve(&alice, 400 * PLMC));
+		assert_ok!(Balances::set_freeze(
+			&polimec_base_runtime::RuntimeFreezeReason::Democracy(pallet_democracy::FreezeReason::Vote),
+			&alice,
+			400 * PLMC
+		));
+		Balances::set_lock(*b"py/trsry", &alice, 400 * PLMC, WithdrawReasons::all());
```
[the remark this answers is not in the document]
vstam1: Its mostly that Reserves after locks do not work. So trying to use a reserve on your vesting scheduled tokens will result in an error. So for example people couldn't propose or second a public proposal with their vesting tokens or couldn't candidate themselves in election as it requires reserve. So thats mainly why we moved away from old traits for democracy and election pallets. Election for example also requires a reserve for voting in the original pallet, but we changed that to a vote locking period now. Another change that we made compared to the old pallets was being able to use total_balance instead of free balance for voting else tokens in staking couldn't be used for voting. But yeah you can place a lock/freeze on  reserved balance but you can't do it the other way around.

lrazovic:
Following our offline discussion, we have agreed to proceed with merging this.

PR #163, opened 2024-02-12
Feature/plmc 432 allow async instantiaton of auctioning project with existing

JuaniRios:
## What?
- We can now include a list of bids to be included in an auctioning project when running the "polimec-testing" node

## Why?
- We wanted to test the weighted average price calculation with max bids

## How?
- We added the same logic we used in the community funding creation function, inside the auction creation function

## Testing?
- new test in pallet-funding written: `starting_auction_round_with_bids`
- run `build_spec_testing_node` from the instantiator

## Anything Else?
- Fixed the issue where we needed to mint the metadata deposit to the issuer in the async functions
- Changed the update block from being calculated from the config types to inspecting the `ProjectsToUpdate`.
  - This is due to the config `MaxProjectsToUpdatePerBlock` set to 1, and so any parallel instantiation won't know exactly when the update block is without inspecting the storage


pallets/funding/src/instantiator.rs
```diff
@@ -1876,10 +1887,11 @@ pub mod async_features {
 
 		let mut inst = instantiator.lock().await;
 
-		match inst.get_project_details(project_id).status {
-			ProjectStatus::RemainderRound | ProjectStatus::FundingSuccessful => Ok(()),
-			_ => panic!("Bad state"),
-		}
+		assert_matches!(
```
vstam1, line 1890: Do we actually need this library for a single assert? Would rather remove the library and have 2-3 lines more code to do this assert manually. Cant we do something like `assert!(matches!(status, Remainder | Funding ))`

pallets/funding/src/lib.rs
```diff
@@ -1535,7 +1535,7 @@ pub mod pallet {
 				<T as Config>::SetPrices::set_prices();
 				instantiator::async_features::create_multiple_projects_at(inst, self.starting_projects.clone());
 
-				frame_system::Pallet::<T>::set_block_number(0u32.into());
+				// frame_system::Pallet::<T>::set_block_number(0u32.into());
```
vstam1, line 1538: Should remove or uncomment

pallets/funding/src/tests.rs
```diff
@@ -6133,4 +6133,68 @@ mod async_tests {
...
+
+			..Default::default()
+		}
+		.assimilate_storage(&mut t)
+		.unwrap();
+
+		let ext = sp_io::TestExternalities::new(t);
+		let mut inst = MockInstantiator::new(Some(RefCell::new(ext)));
+
+		dbg!(inst.get_project_details(0).status);
```
vstam1, line 6194: Can we remove dbg! statements

vstam1 (changes requested):
Lgtm except couple of comments

pallets/funding/src/instantiator.rs
JuaniRios: we use it extensively in tests.rs, so why not use it here too even if we use it once?
I get that if we use it only once throughout the code base then it's silly to download a library for it, but since we already have it in the workspace why not use it and reduce code bloat?

pallets/funding/src/lib.rs
JuaniRios: Done, although it shouldn't have much effect since zombienet sets the block to 0 anyway, but felix had some problems without this line I think

pallets/funding/src/benchmarking.rs
```diff
@@ -1657,7 +1603,9 @@ mod benchmarks {
 			vec![],
 		);
 
-		inst.advance_time(<T as Config>::SuccessToSettlementTime::get()).unwrap();
+		let (settlement_block, _) = inst.get_update_pair(project_id);
```
lrazovic, line 1606: Given that these three lines are repeated multiple times, could we encapsulate them within a function?

pallets/funding/src/functions.rs
```diff
@@ -564,7 +584,7 @@ impl<T: Config> Pallet<T> {
 				project_details.status = ProjectStatus::CommunityRound;
 				ProjectsDetails::<T>::insert(project_id, project_details);
 				// TODO: return real weights
```
lrazovic, line 586: We can remove this TODO

lrazovic (approved):
Just two minor comments, logic wise everything seems fine!

pallets/funding/src/functions.rs
JuaniRios: done

pallets/funding/src/benchmarking.rs
JuaniRios: done

vstam1 (approved):

PR #167, opened 2024-02-14
add extra proxy functionalities

vstam1:
## What?
Adding more proxy types to the base runtime. We add the following proxy types:
- NonTransfer: Should allow all (non-system) pallet calls.
- Governance: All governance related pallets.
- Staking: Only allow calls for ParachainStaking pallet.
- IdentityJudgement: Only allow the provide_judgement call in Identity pallet for Registrars to judge on identities.

## Why?
Each of these proxy types give different permissions to the proxy account. An important potential usecase can be exchanges that want some proxy functionality for their funds. 

## How?
InstanceFilter for proxies give precise access to calls per proxy type.

## Testing?

Thinking about testing it on rolimec. Could write some integration tests but not sure if it actually makes sense.

## Anything Else?

Should we write integration tests?


JuaniRios (approved):

runtimes/base/src/lib.rs
```diff
@@ -242,17 +242,56 @@ impl Contains<RuntimeCall> for BaseCallFilter {
...
+			ProxyType::Governance => matches!(
+				c,
+				RuntimeCall::Treasury(..) |
+					RuntimeCall::Democracy(..) |
+					RuntimeCall::Council(..) |
+					RuntimeCall::TechnicalCommittee(..) |
+					RuntimeCall::Elections(..) |
+					RuntimeCall::Preimage(..) |
+					RuntimeCall::Scheduler(..)
+			),
```
lrazovic, line 279: This formatting looks very strage, but according to `cargo +nightly fmt` it is correct 🤷‍♂️

lrazovic (approved):

PR #168, opened 2024-02-15
Fix voting issue in election pallet with premature unlock of tokens

vstam1:
## What?
Add check on active stake lock in election phragmen pallet while voting. New balance stake should be at least similar to the active lock.

## Why?
Currently someone can circumvent the active voting lock by first voting with a high balance (say 1000 PLMC) just before an election, and then set a new vote just after the election with a low balance (say 1 PLMC), thereby releasing the lock on the balance difference (999 PLMC). 

## How?
We check if there is a currently active lock. If the lock is in place, the balance required for voting has to be equal or higher then the currently active locked balance. 

## Testing?
I adjusted some failing tests to the new behaviour.


lrazovic (approved):

JuaniRios (approved):
lgtm, nice description

PR #170, opened 2024-02-19
Feature/on chain jwt

vstam1:


polimec-common/common/src/credentials/mod.rs
```diff
@@ -0,0 +1,136 @@
...
+						}
+
+						if claims.custom.investor_type == Type::get() && claims.custom.subject == who {
+							return Ok(who);
+						}
+					}
+				}
+			}
+		}
+		Err(origin)
```
JuaniRios, line 92:
```suggestion
		if let Ok(claims) = Self::verify_token(token, verifying_key)
			&& let Some(who) = origin.clone().into_signer()
			&& let Ok(now) = Now::<Timestamp>::get().try_into()
			&& let Some(date_time) = claims.expiration
			&& claims.custom.investor_type == Type::get()
			&& claims.custom.subject == who
			&& date_time.timestamp() as u64 >= now {
				return Ok(who);
			}

		Err(origin)
```

polimec-common/common/src/lib.rs
```diff
@@ -18,8 +18,8 @@
```
JuaniRios, line 18:
```suggestion
#![feature(let_chains)]
```

JuaniRios (approved):
LGTM.
We could use if-let chains to make the code cleaner.
If so we should replace the toolchain to something like "nightly-2023-12-01" which is confirmed to be working on my machine.
I'm already using that toolchain always locally since my indexing doesn't work with stable due to substrate crates using nightly features when building

lrazovic (approved):
Congratulations, minimal yet functional implementation, very good work indeed!

Thanks also to @JuaniRios  for giving the idea of the various custom "EnsureOrigin"!

PR #172, opened 2024-02-20
Increase ED by 10x. 0.001 PLMC -> 0.01 PLMC

JuaniRios:
## What?
Increase PLMC ED by 10x

## Why?
Not sure, @lrazovic could comment on that

## How?
Change the const in the shared-configuration

## Testing?
Run all tests and check they didn't break


vstam1 (approved):
PR looks good to me if necessary. Let @lrazovic judge that

lrazovic (approved):
LGTM. 

A `pre_upgrade()` check using the `try-runtime-cli` showed that no account will be killed in this way.

PR #174, opened 2024-02-21
feat: add new migration

lrazovic:


JuaniRios (approved):
Migration ran ok on my machine. Code looks good, but can't judge the accuracy of the accounts. Weights also look good

PR #176, opened 2024-02-23
Feature/split contributions

vstam1:
## What?
- Split extrinsics into community and remainder contributions
- Simplify contribute benchmarks and weight returns
- call inner do_functions in tests

## Why?
- In the community, we need to check if the user has a winning bid already, and if so return err. The credential check should be the same though.
  - The token amount left to buy is different in each round
- Was too complex, making it harder to maintain. It's now less accurate but its a worthy tradeoff
- Extrinsic calls will do credential checks, and we don't want those in pallet tests. We do those in integration tests 

## How?
- abstract common logic into private function, create 2 public functions with a distinct logic
- assume we always create the ct account, and we always do some sort of iteration over users. We now have only 2 logic paths, either the contribution ends the round or it doesn't. The rest of the logic branches get included in the linear interpolation (even though it's not a linear relationship)

## Testing?
- Since we made the original do_contribute private, we are now forced to replace all calls with either community or remainder contribute calls. So just run `cargo test`
- Run benchmark tests and benchmark weight generation and check results.
- Add a test for the winning bidder to not be accepted in the community but in the remainder.

## Anything Else?
Next PR should make the default values used in tests to be more complex, to be in line with the new logic of who gets to participate in the community.


JuaniRios (approved):
I coded the last part of the PR so LGTM

integration-tests/src/tests/basic_comms.rs
lrazovic, line 1: What's the reasoning behind these changes?

integration-tests/src/tests/mod.rs
```diff
@@ -14,6 +14,7 @@
 // You should have received a copy of the GNU General Public License
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
+// infinite loop
```
lrazovic, line 17:
```suggestion
```

pallets/funding/src/weights_old.rs
lrazovic, line 1: Can we remove this file?

rust-toolchain.toml
```diff
@@ -1,4 +1,4 @@
 [toolchain]
-channel = "1.74.0"
+channel = "nightly-2023-12-01"
```
lrazovic, line 2:
```suggestion
channel = "1.74.0"
```

lrazovic (changes requested):

integration-tests/src/tests/basic_comms.rs
JuaniRios: oops, forgot to uncomment

pallets/funding/src/weights_old.rs
JuaniRios: yes

integration-tests/src/tests/basic_comms.rs
```diff
@@ -20,6 +20,9 @@ use crate::*;
 const MAX_REF_TIME: u64 = 300_000_000;
 const MAX_PROOF_SIZE: u64 = 10_000;
 
+#[test]
+fn test() {}
+
```
lrazovic, line 25: Is this needed?

lrazovic (approved):
One last very minor comment, but great simplification!

integration-tests/src/tests/basic_comms.rs
JuaniRios: good catch!

PR #180, opened 2024-03-06
Feature/plmc 442 delete remove from update store

vstam1:
## What?
This pr deletes the `remove_from_update_store` function. This means that we do not remove any project transitions from the update store anymore, if these transitions are not necessary anymore. Instead we handle these unnecessary transitions by gracefully failing the second execution of these transitions. These double transition execution happen, as the project issuer can manually push a project on, or in the situation from the community round, where the round is filled during the community round duration. There were 3 cases of removal of the automatic transition:

- do_english_auction: Auction can be started manually by the Issuer. We now fail the second do_english_auction by checking if the project is in the `AuctionInitializePeriod` status.
- do_contribute: Community round was filled, so we end the funding. Automatic end of funding is now still scheduled in the update store. We fail the second funding by checking if the project status is already either of the following options:
`ProjectStatus::FundingSuccessful | ProjectStatus::FundingFailed | ProjectStatus::AwaitingProjectDecision`
- do_decide_project_outcome: In case a project is not directly failed or successful [70%, 90%], the project issuer can chose to accept or reject successful funding. In the second execution of `do_project_decision` we check if the project is still in the `AwaitingProjectDecision` phase. 

## Why?
To simplify the benchmarking complexity by removing the removal from project store attempts. 

## Testing?
Fixed testing by updating the get_update_pair to take an UpdateType. As multiple round transition can be stored in update store for a project, and the order of those transitions is not always clear, we now have to explicitly specify which transition we are expecting. This is the case for the funding tests, benchmarks and the integration tests.

### DIFF
--- a/pallets/funding/src/functions.rs
+++ b/pallets/funding/src/functions.rs
@@ -608,7 +608,7 @@ impl<T: Config> Pallet<T> {
 		ensure!(project_details.status == ProjectStatus::CommunityRound, Error::<T>::ProjectNotInCommunityRound);
 
 		// Transition to remainder round was initiated by `do_community_funding`, but the ct
-		// tokens where already sold in the community round. This transition is obsolite.
+		// tokens where already sold in the community round. This transition is obsolete.
 		ensure!(
 			project_details
 				.remaining_contribution_tokens
--- a/pallets/funding/src/functions.rs
+++ b/pallets/funding/src/functions.rs
@@ -610,7 +610,11 @@ impl<T: Config> Pallet<T> {
 		// Transition to remainder round was initiated by `do_community_funding`, but the ct
 		// tokens where already sold in the community round. This transition is obsolite.
 		ensure!(
-			project_details.remaining_contribution_tokens.0.saturating_add(project_details.remaining_contribution_tokens.1) > 0u32.into(),
+			project_details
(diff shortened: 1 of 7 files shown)


pallets/funding/src/functions.rs
```diff
@@ -639,6 +607,17 @@ impl<T: Config> Pallet<T> {
 		ensure!(now > community_end_block, Error::<T>::TooEarlyForRemainderRoundStart);
 		ensure!(project_details.status == ProjectStatus::CommunityRound, Error::<T>::ProjectNotInCommunityRound);
 
+		// Transition to remainder round was initiated by `do_community_funding`, but the ct
+		// tokens where already sold in the community round. This transition is obsolite.
```
JuaniRios, line 611: "obsolete"

JuaniRios (changes requested):
Some small things and we can merge

lrazovic (commented):
Always happy to see simplifications wherever possible. Nothing to add beyond Juan's comments.

JuaniRios:
I talked to @vstam1 and decided to finish the PR for him since he is working on #181.
@lrazovic can you guys take a last look and we merge?

lrazovic (approved):

JuaniRios (approved):

PR #181, opened 2024-03-07
Feature/plmc 460 upgrade to sdk 16

vstam1:
## What?
Upgrade Polimec to [Polkadot sdk 1.6](https://github.com/paritytech/polkadot-sdk/releases/tag/polkadot-v1.6.0)

Checklist:
- [x] Make everything compile
- [x] Make all tests pass
- [x] Make benchmarks pass
    - [ ] Run all benchmarks
- [x] Check if upgrade would pass with try-runtime (has bug but pallet version upgrades seem to pass)
    - try-runtime has a bug with decoding HostConfig, see [stackexchange post](https://substrate.stackexchange.com/questions/10986/runtime-upgrade-for-parachainsystemhostconfiguration) 
- [x] Update xcm for the new pallet-xcm implementation
- [x] run fmt

### DIFF
--- a/justfile
+++ b/justfile
@@ -3,11 +3,11 @@ default:
     @just --list
 
 # Build the "Base" Runtime using srtool
-build-base-srtool:
-    srtool build --root -p polimec-runtime --runtime-dir runtimes/base --build-opts="--features=on-chain-release-build"
+build-polimec-srtool:
+    srtool build --root -p polimec-runtime --profile production --runtime-dir runtimes/polimec --build-opts="--features=on-chain-release-build" --no-wasm-std
 
 build-rolimec-srtool:
-    srtool build --root -p polimec-runtime --runtime-dir runtimes/base --build-opts="--features=on-chain-release-build,fast-mode"
+    srtool build --root -p polimec-runtime --profile production --runtime-dir runtimes/polimec --build-opts="--features=on-chain-release-build,fast-mode" --no-wasm-std
 
 # Build the "Testnet" Runtime using srtool
 
@@ -20,10 +20,10 @@ test-integration:
     cargo test -p integration-tests
 
 # src: https://github.com/polkadot-fellows/runtimes/blob/48ccfae6141d2924f579d81e8b1877efd208693f/system-parachains/asset-hubs/asset-hub-polkadot/src/weights/cumulus_pallet_xcmp_queue.rs
-# Benchmark a specific pallet on the "Base" Runtime
+# Benchmark a specific pallet on the "Polimec" Runtime
 # TODO: Adjust the `--chain` flag to match the chain you are benchmarking
-benchmark-runtime chain="base-rococo" pallet="pallet-elections-phragmen" features="runtime-benchmarks":
-    cargo run --features {{ features }} --release -p polimec-parachain-node benchmark pallet \
+benchmark-runtime chain="polimec-local" pallet="pallet-elections-phragmen" features="runtime-benchmarks":
+    cargo run --features {{ features }} --profile production -p polimec-node benchmark pallet \
       --chain={{ chain }} \
       --steps=50 \
       --repeat=20 \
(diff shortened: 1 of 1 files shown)


pallets/linear-release/src/mock.rs
```diff
@@ -124,6 +97,9 @@ pub enum MockRuntimeHoldReason {
 	Reason,
 	Reason2,
 }
+impl VariantCount for MockRuntimeHoldReason {
```
JuaniRios, line 100: Can't we just derive `VariantCount` from the crate `variant_count`, which always gives us the correct number of variants at compile time?
Or is that trait necessary for HoldReason?

JuaniRios: I'm already using that crate for `AcceptedFundingAsset`

JuaniRios (changes requested):
The integration tests are broken, once they are fixed I'm ok with merging this.

There are a lot of changes to config types of pallets like frame-system, pallet-xcm, and the like, and I assume you took the default values used by polkadot, so I think we're safe in that regard.

On the emulator side, as long as the tests pass, then the new config you setup will be fine.

Our xcm config didn't seem to change much, except in replacing the transaction from using currency to fungible.

The new task system is now added, so it might be good to start tinkering with it and maybe replace our `Cleaner` implementation with it.

I also saw there is a bridge mentioned in the emulator, so does that mean we can test a future ct migration to something like ethereum in the emulator?

pallets/funding/src/functions.rs
```diff
@@ -2141,23 +2133,17 @@ impl<T: Config> Pallet<T> {
...
+				let sender_id = ParaId::from(sender).encode();
+				accept_channel_relay_call.extend_from_slice(&sender_id);
+
+				let mut request_channel_relay_call = vec![60u8, 0];
+				let recipient = ParaId::from(sender).encode();
+				request_channel_relay_call.extend_from_slice(&recipient);
+				let proposed_max_capacity = T::RequiredMaxCapacity::get().encode();
+				request_channel_relay_call.extend_from_slice(&proposed_max_capacity);
+				let proposed_max_message_size = T::RequiredMaxMessageSize::get().encode();
+				request_channel_relay_call.extend_from_slice(&proposed_max_message_size);
```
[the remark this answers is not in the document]
vstam1: We still have a dependency in the integration tests. However, integration tests are of course seperate to our node building so we should not have any dep on polkadot runtime in the node anymore.

runtimes/shared-configuration/src/identity.rs
```diff
@@ -14,14 +14,16 @@
...
 use frame_support::parameter_types;
 
 parameter_types! {
 	/// The basic deposit to create an identity.
 	pub const BasicDeposit: Balance = 20 * PLMC;
 	/// Deposit for each additional field.
-	pub const FieldDeposit: Balance = 200 * MILLI_PLMC;
+	pub const ByteDeposit: Balance = deposit(0, 1);
+
+	pub const PendingUsernameExpiration: u32 = 0;
```
lrazovic, line 26: I would add a short comment to be aligned with the rest

lrazovic (changes requested):
Nice! I align myself with Juan's comments - as soon as the e2e tests are settled we can merge

pallets/linear-release/src/mock.rs
vstam1: Don't think this works for RuntimeHoldReason as there is a frame trait bound and this crate doesn't implement this specific trait as its frame specific.

vstam1:
@lrazovic @JuaniRios Integrations should be fixed and changed the cargo toml of funding pallet.

JuaniRios:
- `cargo build --release --features std,fast-mode` is not compiling correctly. I'll fix it.
- `from_genesis` is deprecated, we should probably use `builder()`

JuaniRios (approved):
Benchmarks were failing on the testnet runtime. Pushed some commits to fix them.
We can merge it once we figure out the node sync issuer @lrazovic mentioned

lrazovic (approved):

PR #183, opened 2024-03-11
Feature/plmc 439 liquidity pool and long term holder bonus rewards minting

JuaniRios:
Change base to main after #177 is merged.

## What?
- Add new contribution token treasury
- Add minting of long-term holder bonus, and liquidity pool rewards to the contribution token treasury

## Why?
- Functionality defined in the Knowledge Hub

## How?
- New config type for the new treasury account
- After creating the contribution token, calculate the amounts and mint them. 

## Testing?
- one test called `ct_treasury_mints`


pallets/funding/src/functions.rs
```diff
@@ -3164,6 +3179,71 @@ impl<T: Config> Pallet<T> {
...
+		let project_details = ProjectsDetails::<T>::get(project_id).ok_or(Error::<T>::ProjectNotFound)?;
+		let project_metadata = ProjectsMetadata::<T>::get(project_id).ok_or(Error::<T>::ProjectNotFound)?;
+
+		// Determine how much funding has been achieved.
+		let funding_amount_reached = project_details.funding_amount_reached;
+		let fundraising_target = project_details.fundraising_target;
+		let total_issuer_fees = Self::calculate_fees(funding_amount_reached);
+
+		let initial_token_allocation_size = project_metadata.total_allocation_size;
+		let final_remaining_contribution_tokens = project_details.remaining_contribution_tokens;
```
lrazovic, line 3193: The preamble of these two new functions (`generate_long_term_holder_reward_amount` and `generate_liquidity_pools_reward_amount`) looks like identical, and they are called one after the other. Do we actually have to compute these values fresh every time, or can we cache the compution made by the first function and pass it to the second?

pallets/funding/src/mock.rs
```diff
@@ -414,13 +415,14 @@ construct_runtime!(
...
+	let ed = <TestRuntime as pallet_balances::Config>::ExistentialDeposit::get();
 	RuntimeGenesisConfig {
 		balances: BalancesConfig {
-			balances: vec![(
-				<TestRuntime as Config>::PalletId::get().into_account_truncating(),
-				<TestRuntime as pallet_balances::Config>::ExistentialDeposit::get(),
-			)],
+			balances: vec![
+				(<TestRuntime as Config>::PalletId::get().into_account_truncating(), ed),
+				(<TestRuntime as Config>::ContributionTreasury::get(), ed),
```
lrazovic, line 423: Curiosity: why should the Contribution Treasury have at least ED?

lrazovic (changes requested):
Glad to see that we are getting closer and closer to having all the features described in the Knowledge Hub! I left some questions on the technical implementation of the new logic.

pallets/funding/src/mock.rs
JuaniRios: Good catch!
I assumed we could just use the PLMC ED to use a provider for a new pallet-assets account for that token, but we would run into the same problem we had before with regular participants.

So someone needs to pay for the account deposit by calling "touch". Should the issuer pay for it or the PLMC treasury?

pallets/funding/src/functions.rs
JuaniRios: Yea I chose readability over DRY on this one. If you don't like it I can have one function returning both values or like you said have the values precomputed and passed to the functions

pallets/funding/src/functions.rs
```diff
@@ -872,6 +873,20 @@ impl<T: Config> Pallet<T> {
 				token_information.decimals,
 			)?;
 
+			let long_term_holder_bonus_ct_amount = Self::generate_long_term_holder_reward_amount(project_id)?;
```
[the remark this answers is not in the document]
JuaniRios: yea the storage layers should work fine here, so we won't do another read. It makes it easier to read this way than passing everything as variables when we don't need to

pallets/funding/src/functions.rs
lrazovic: Ok, but what about the `total_issuer_fees`? Isn't that computed two times by executing `calculate_fees`?

pallets/funding/src/mock.rs
lrazovic: Should this be a one-off thing or should the account have enough PLMC for each project?

vstam1 (commented):
Some small remarks + Leo's comments about the two functions that could be combined.

pallets/funding/src/functions.rs
JuaniRios: Alright, I'm going to combine both functions into one, since we call them always together

pallets/funding/src/mock.rs
JuaniRios: Every new asset you want to transfer/mint to an account with pallet-assets, requires either increasing the provider count of the account (which you are limited to), or paying for the storage of that asset.

System accounts behave the same as regular accounts, so they need an ED for creating the account, and they have the provider count increased by other pallets like pallet-assets.

This won't be a one-off thing, since each project will have a different asset in pallet-assets, so will need a new provider count, or a new payment for the storage creation. 

I will make the issuer pay for it

runtimes/testnet/src/lib.rs
```diff
@@ -650,7 +650,7 @@ impl pallet_funding::Config for Runtime {
 	type CandleAuctionDuration = CandleAuctionDuration;
 	type CommunityFundingDuration = CommunityFundingDuration;
 	type ContributionTokenCurrency = LocalAssets;
-	type ContributionVesting = ContributionVestingDuration;
+	type ContributionTreasury = TreasuryAccount;
```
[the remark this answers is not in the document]
JuaniRios: No clue. Something to be discussed when we move this to base runtime

pallets/funding/src/functions.rs
```diff
@@ -134,6 +134,19 @@ impl<T: Config> Pallet<T> {
...
+		// We make the issuer pay that deposit for the contribution token treasury account.
+		let contribution_token_treasury_account = <T as Config>::ContributionTreasury::get();
+		let deposit = T::ContributionTokenCurrency::deposit_required(project_id);
+		T::NativeCurrency::transfer(issuer, &contribution_token_treasury_account, deposit, Preservation::Preserve)
+			.map_err(|_| Error::<T>::NotEnoughFundsForCTDeposit)?;
+		T::NativeCurrency::hold(
+			&HoldReason::FutureDeposit(project_id).into(),
+			&contribution_token_treasury_account,
+			deposit,
+		)
```
vstam1, line 147: Is this hold actually necessary? These tokens won't go anywhere, so I don't think we need to "hold" them?

pallets/funding/src/functions.rs
```diff
@@ -849,6 +863,34 @@ impl<T: Config> Pallet<T> {
...
 			)?;
 
+			let contribution_token_treasury_account = T::ContributionTreasury::get();
+			let deposit = T::ContributionTokenCurrency::deposit_required(project_id);
+			T::NativeCurrency::release(
+				&HoldReason::FutureDeposit(project_id).into(),
+				&contribution_token_treasury_account,
+				deposit,
+				Precision::BestEffort,
+			)?;
```
vstam1, line 873: Should the tokens be "released" (if necessary if we keep the hold) and returned to the issuer on project_failure?

pallets/funding/src/functions.rs
JuaniRios: Just to be super sure that no one is able to spend those PLMC on something else by accident, even If the only way to spend them is with governance

pallets/funding/src/functions.rs
JuaniRios: You're right

pallets/funding/src/functions.rs
vstam1: Might be a bit overkill. If this "treasury" account is different from the treasury account of the treasury pallet, these tokens would only be able to be spend by governance. But that would be a potentially malicious proposal, so should be blocked by council/TC. If we already talking about malicious proposal, it could probably already include something to remove hold as well. So lets keep it simpel and do not add an (in my eyes) unnecessary hold.

lrazovic (approved):
As soon as the open discussions are resolved, we can go ahead and merge this.

PR #193, opened 2024-03-18
Feat/send dust to treasury

lrazovic:
## What?

When an account is closed but still holds a balance below the Existential Deposit (ED), some PLMC "dust" is generated. This dust should be directed to the Blockchain Operation Treasury instead of being burned.

## Why?

The following extrinsic https://explorer.polimec.org/polimec/extrinsic/767479-2 resulted in the account of the sender being killed, and the residual dust was burned rather than allocated to the Blockchain Operation Treasury.

## How?

We can adjust this behavior through the `DustRemoval` configuration option in the `Balances` pallet.

## Testing?

A new integration test now compares the total issuance before and after a `DustLost` event to ensure the dust is properly accounted for.

### DIFF
--- a/integration-tests/src/constants.rs
+++ b/integration-tests/src/constants.rs
@@ -469,6 +469,7 @@ pub mod polimec_base {
 	use super::*;
 	use crate::PolimecBase;
 	use pallet_funding::AcceptedFundingAsset;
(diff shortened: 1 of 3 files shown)


JuaniRios (approved):
lgtm

vstam1 (approved):

PR #195, opened 2024-03-18
Feature/plmc 467 return deposits to issuer on project fail

JuaniRios:
## What?
- We charge the issuer PLMC for the ct account of the ct-treasury, the metadata deposit of the ct, and the ED of the escrow.
  - If the project fails, then no CT is created, therefore we can return the deposits associated with it.
  - The escrow still needs to live since it holds users' funds that need to be returned.

## Why?
- We weren't returning them before

## How?
- In the same place we were creating the CT on project success, we return CT deposits on project fail.
  - This is the function that started the `Cleaner`.

## Testing?
- New function called `creation_deposits_returned_on_funding_fail`


vstam1:
Probably not necessary anymore after deleting CT requirement

PR #199, opened 2024-03-19
feature/plmc-468-limit-multipliers-based-on-credential-type

JuaniRios:
## What?
- Limit what multiplier participants can specify based on their credential

## Why?
- Knowledge hub requirement

## How?
set constants for the constraints in types.rs file, and check on the participation functions if it's allowed

## Testing?
`non_retail_multiplier_limits` and `retail_multiplier_limits` functions on each round module

## Anything Else?
Also check that its > 0


pallets/funding/src/functions.rs
```diff
@@ -1309,7 +1332,20 @@ impl<T: Config> Pallet<T> {
...
 		};
+		let max_multiplier = match investor_type {
+			InvestorType::Retail => {
+				RetailParticipations::<T>::mutate(&did, |project_participations| {
+					if project_participations.contains(&project_id).not() {
+						// We don't care if it fails, since it means the user already has access to the max multiplier
+						let _ = project_participations.try_push(project_id);
+					}
+					retail_max_multiplier_for_participations(project_participations.len() as u8)
+				})
```
lrazovic, line 1343: nit: maybe a [`try_mutate`](https://paritytech.github.io/polkadot-sdk/master/frame_support/storage/types/struct.StorageMap.html#method.try_mutate) where we modify the `RetailParticipations` is more suitable?

lrazovic (changes requested):
It looks good! Just few nits here and there

pallets/funding/src/mock.rs
```diff
@@ -398,9 +398,9 @@ impl Config for TestRuntime {
 	type InvestorOrigin = EnsureInvestor<TestRuntime>;
 	type ManualAcceptanceDuration = ManualAcceptanceDuration;
 	type MaxBidsPerProject = ConstU32<1024>;
-	type MaxBidsPerUser = ConstU32<4>;
+	type MaxBidsPerUser = ConstU32<150>;
```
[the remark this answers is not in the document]
JuaniRios: I needed it this high to test a successful bid for each multiplier 1-20, and unsuccessful for higher ones until 50. We can probably reduce it to 70 or 80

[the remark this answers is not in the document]
JuaniRios: Actually, since for multiplier 21-50 it will fail, there's no need to have it higher than 20 or 21

pallets/funding/src/mock.rs
```diff
@@ -398,9 +398,9 @@ impl Config for TestRuntime {
...
 	type MaxCapacityThresholds = MaxCapacityThresholds;
-	type MaxContributionsPerUser = ConstU32<4>;
+	type MaxContributionsPerUser = ConstU32<50>;
```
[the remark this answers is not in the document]
JuaniRios: Here I need it high too to test all multipliers for retail from 1 to 10x, for the 5 tiers 	
(past_participations,max_multiplier): (2, 1),(4, 2),(9, 4),(24, 7),(25, 10).

pallets/funding/Cargo.toml
```diff
@@ -54,6 +54,7 @@ polimec-common-test-utils.workspace = true
 macros.workspace = true
 xcm-builder.workspace = true
 xcm-executor.workspace = true
+env_logger = "0.10.2"
```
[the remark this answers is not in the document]
JuaniRios: env_logger helps set up the logging environment. You cannot use log directly on tests without having some logging runtime activated

pallets/funding/src/types.rs
```diff
@@ -154,6 +164,20 @@ pub mod config_types {
...
 	}
+
+	pub type MaxParticipationsForMaxMultiplier = ConstU32<25>;
+	pub const fn retail_max_multiplier_for_participations(participations: u8) -> u8 {
+		match participations {
+			0..=2 => 1,
+			3..=4 => 2,
+			5..=9 => 4,
+			10..=24 => 7,
+			25..=u8::MAX => 10,
```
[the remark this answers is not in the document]
JuaniRios: Why would the max be a constant and not the other tiers then, if none of them follow any logic?
I would either use cont on all of them, or no consts at all no?

pallets/funding/src/functions.rs
JuaniRios: no, because this function always need to return a value, and try_mutate expects a `None` if I want to not change the value. So I would need to always return `Some` which would do the same as `mutate`

pallets/funding/src/types.rs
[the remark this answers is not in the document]
lrazovic: Ah sorry, I thought the 10 was a typo. Then `RETAIL_MAX_MULTIPLIER` is unused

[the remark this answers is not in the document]
JuaniRios: Good catch 🙏🏻

pallets/funding/src/mock.rs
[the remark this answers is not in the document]
JuaniRios: You're right, and for bids it was a matter of putting the multiplier check before the total count check

pallets/funding/src/mock.rs
```diff
@@ -398,9 +398,9 @@ impl Config for TestRuntime {
 	type InvestorOrigin = EnsureInvestor<TestRuntime>;
 	type ManualAcceptanceDuration = ManualAcceptanceDuration;
 	type MaxBidsPerProject = ConstU32<1024>;
-	type MaxBidsPerUser = ConstU32<4>;
+	type MaxBidsPerUser = ConstU32<25>;
```
vstam1, line 401: I discussed with Luca that we will limit the MaxBidsPerUser and MaxContributionsPerUser to 16.

pallets/funding/src/functions.rs
```diff
@@ -946,12 +947,19 @@ impl<T: Config> Pallet<T> {
...
 
 		// * Calculate new variables *
+		if investor_type == InvestorType::Retail {
+			RetailParticipations::<T>::mutate(&did, |project_participations| {
+				if project_participations.contains(&project_id).not() {
+					// We don't care if it fails, since it means the user already has access to the max multiplier
+					let _ = project_participations.try_push(project_id);
+				}
+			});
+		}
```
vstam1, line 957: I could not find anything specific on knowledge hub about evaluation counting as "participating in funding round". Are we sure a retail investor has its number increased if he/she evaluates

pallets/funding/src/functions.rs
```diff
@@ -1064,6 +1072,20 @@ impl<T: Config> Pallet<T> {
...
+			InvestorType::Retail => {
+				RetailParticipations::<T>::mutate(&did, |project_participations| {
+					if project_participations.contains(&project_id).not() {
+						// We don't care if it fails, since it means the user already has access to the max multiplier
+						let _ = project_participations.try_push(project_id);
+					}
+					retail_max_multiplier_for_participations(project_participations.len() as u8)
+				})
+			},
+
```
vstam1, line 1085: Retail cannot bid, right? So why do we have a check on max_multiplier here? We return with error two lines above this statement if type is Retail

pallets/funding/src/functions.rs
```diff
@@ -1309,8 +1332,22 @@ impl<T: Config> Pallet<T> {
...
 		};
+		let max_multiplier = match investor_type {
+			InvestorType::Retail => {
+				RetailParticipations::<T>::mutate(&did, |project_participations| {
+					if project_participations.contains(&project_id).not() {
+						// We don't care if it fails, since it means the user already has access to the max multiplier
+						let _ = project_participations.try_push(project_id);
+					}
+					retail_max_multiplier_for_participations(project_participations.len() as u8)
+				})
```
vstam1, line 1343: If this "participation" only counts if a Retail investor participates during contribution round and not evaluation, we could do a check on the `ContributionBoughtUSD` if balance == 0 for (project_id, did) to determine if its the first participation for this project_id. This way we do not have to store a full vec of project_ids that retail investor participated in, but just a u32 or something.

pallets/funding/src/functions.rs
```diff
@@ -1309,8 +1332,22 @@ impl<T: Config> Pallet<T> {
...
+			},
 
+			InvestorType::Professional => PROFESSIONAL_MAX_MULTIPLIER,
+			InvestorType::Institutional => INSTITUTIONAL_MAX_MULTIPLIER,
+		};
 		// * Validity checks *
+		ensure!(multiplier.into() <= max_multiplier && multiplier.into() > 0u8, Error::<T>::ForbiddenMultiplier);
```
vstam1, line 1350: We have a bug in the do_contribute_(community/remainder)_round functions as we are not limiting the types of investors anymore. We just pass the investor type to the inner function. But Institutional (and professional?) cannot invest in community round.

vstam1 (changes requested):

pallets/funding/src/mock.rs
JuaniRios: Ok, but the mock needs more for the multiplier tests

pallets/funding/src/functions.rs
JuaniRios: Luca said yes

pallets/funding/src/functions.rs
JuaniRios: Good point lol

pallets/funding/src/functions.rs
JuaniRios: Institutional and professional CAN invest in community round. As long as they have no winning bids

vstam1 (approved):

lrazovic (approved):

PR #200, opened 2024-03-20
Feature/settlement rewrite

vstam1:

### DIFF
--- a/pallets/funding/src/lib.rs
+++ b/pallets/funding/src/lib.rs
@@ -1625,19 +1625,71 @@ pub mod pallet {
 
 			// We want at least 5% of max block weight
 			let at_least = MAXIMUM_BLOCK_WEIGHT.saturating_div(20);
-			if remaining_weight.all_gt(at_least)  {
+			if !remaining_weight.any_lt(at_least)  {
 				return Weight::zero();
 			};
 
 			if let Some(project_id) = ProjectSettlementQueue::<T>::get() { 
 				if let Some(project_details) = ProjectsDetails::<T>::get(project_id) {
-					let eval_iter = Evaluations::<T>::iter_key_prefix((project_id,));
-					while remaining_weight.any_gt(at_least) {
-						
-						match project_details.status {
-							ProjectStatus::FundingSuccessful => 
+					let mut eval_iter = Evaluations::<T>::iter_prefix((project_id,));
+					
+					while let Some((_, evaluation)) = eval_iter.next() {
+						if remaining_weight.any_lt(at_least) {
+							return Weight::zero();
 						}
+						let _ = match project_details.status {
+							ProjectStatus::FundingSuccessful => {
+								Self::do_settlement_success_evaluator(evaluation, project_id)
+							},
+							ProjectStatus::FundingFailed | ProjectStatus::EvaluationFailed => {
+								Self::do_settlement_failure_evaluator(evaluation, project_id)
+							},
+							_ => {
+								log::error!("Project status not allowed for settlement: {:?}", project_details.status);
+								Ok(())
+							}
+						};
 					}
+
+					let mut bid_iter = Bids::<T>::iter_prefix((project_id,));
+					
+					while let Some((_, bid)) = bid_iter.next() {
+						if remaining_weight.any_lt(at_least) {
+							return Weight::zero();
+						}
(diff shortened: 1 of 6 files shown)


pallets/funding/src/lib.rs
```diff
@@ -554,6 +560,9 @@ pub mod pallet {
 		ValueQuery,
 	>;
 
+	#[pallet::storage]
+	pub type ProjectSettlementQueue<T: Config> = StorageValue<_, BoundedVec<ProjectId, T::MaxProjectsToSettle>, ValueQuery>;
```
JuaniRios, line 564: Maybe use a `WeakBoundedVec` so even if the queue is full, you can still append stuff. Because what would you do if at funding end the appending fails?

pallets/funding/src/lib.rs
```diff
@@ -1618,48 +1629,128 @@ pub mod pallet {
...
-				})
-				.collect::<Vec<_>>();
-
-			let projects_amount = projects_needing_cleanup.len() as u64;
-			if projects_amount == 0 {
-				return max_weight;
-			}
-
-			let mut max_weight_per_project = remaining_weight.saturating_div(projects_amount);
+			// We want at least 5% of max block weight
```
JuaniRios, line 1632: Are guesstimating a single participation settlement will occupy max 5%?

pallets/funding/src/lib.rs
```diff
@@ -1618,48 +1629,128 @@ pub mod pallet {
...
+			let at_least = MAXIMUM_BLOCK_WEIGHT.saturating_div(20);
+			if remaining_weight.any_lt(at_least)  {
+				return Weight::zero();
+			};
+			let project_ids: BoundedVec<ProjectId, T::MaxProjectsToSettle> = ProjectSettlementQueue::<T>::get();
+			if let Some(&project_id) = project_ids.first()  {
```
JuaniRios, line 1638: use the pattern let Some() = x  else {return} to avoid too many indentations

pallets/funding/src/settlement.rs
```diff
@@ -0,0 +1,306 @@
...
+    pub fn do_settlement_success_bidder(bid: BidInfoOf<T>, project_id: ProjectId) -> DispatchResult {
+        
+        let project_details = ProjectsDetails::<T>::get(project_id).ok_or(Error::<T>::ProjectDetailsNotFound)?;
+		// Ensure that:
+        // 1. The project is in the FundingSuccessful state
+        // 2. The bid is in the Accepted or PartiallyAccepted state
+        // 3. The contribution token exists
+        ensure!(project_details.status == ProjectStatus::FundingSuccessful, Error::<T>::NotAllowed);
+        ensure!(matches!(bid.status, BidStatus::Accepted | BidStatus::PartiallyAccepted(..)), Error::<T>::NotAllowed);
+        ensure!(T::ContributionTokenCurrency::asset_exists(project_id), Error::<T>::CannotClaimYet);
```
JuaniRios, line 32: Just to be aligned, this check is really for making sure settlements only happen after `T::SuccessToSettlementTime`. We check the CT specifically since it is created when the settlement starts.

pallets/funding/src/settlement.rs
```diff
@@ -0,0 +1,306 @@
...
+        T::Vesting::add_release_schedule(
+            &bidder,
+            vest_info.total_amount,
+            vest_info.amount_per_block,
+            funding_end_block,
+            HoldReason::Participation(project_id).into(),
+        )?;
+
+        // Mint the contribution tokens
+        Self::mint_ct_tokens(project_id, &bidder, bid.final_ct_amount)?;
```
JuaniRios, line 50: are we not using the functions associated with the extrinsics? like `do_bid_ct_mint_for`?

JuaniRios (commented):
Just some early feedback

pallets/funding/src/settlement.rs
vstam1: All these functions had a lot of overlapping code so I changed nearly 800 lines of function code to around 300.

pallets/funding/src/lib.rs
vstam1: Its an Guesstimation for now, after creating/running benchmarks we will know more and can adjust this value.

PR #201, opened 2024-03-20
Feature/plmc 472 remove ct deposit fix bug with hold and reserve

vstam1:
## What?
We are removing the ct deposit for the creation of the ct account in pallet_assets

## Why?
As the ct token potentially has a bug where first the deposit could be held while to tokens are locked, but once we actually want to create the account, pallet assets requires a reserve, that does not work together with locked tokens. Upon reviewing the code, we came to the conclusion that we could completely remove the ct tokens and filter out all pallet calls to prevent spamming the chain with empty accounts. 

## How?
Remove all ct deposits

## Testing?
Updated all tests

### DIFF
--- a/runtimes/shared-configuration/src/assets.rs
+++ b/runtimes/shared-configuration/src/assets.rs
@@ -32,7 +32,7 @@ parameter_types! {
 	pub const MetadataDepositBase: Balance = deposit(1, 68);
 	pub const MetadataDepositPerByte: Balance = deposit(0, 1);
 	pub const AssetAccountDeposit: Balance = deposit(1, 18);
-	pub const ZeroAssetAccountDeposit: Balance = 0;
+	pub const ZeroDeposit: Balance = 0;
 }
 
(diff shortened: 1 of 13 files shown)


runtimes/shared-configuration/src/assets.rs
```diff
@@ -32,6 +32,7 @@ parameter_types! {
 	pub const MetadataDepositBase: Balance = deposit(1, 68);
 	pub const MetadataDepositPerByte: Balance = deposit(0, 1);
 	pub const AssetAccountDeposit: Balance = deposit(1, 18);
+	pub const ZeroAssetAccountDeposit: Balance = 0;
```
JuaniRios, line 35: We should also change the `MetadataDepositBase` to 0. Especially since you did it in the mock

JuaniRios (changes requested):

JuaniRios (approved):
lgtm

lrazovic (approved):
Let's fix the warnings in the tests (you can use `cargo fix --lib -p pallet-funding --tests`) and we can merge

PR #207, opened 2024-03-22
Feature/plmc 473 change edit metadata extrinsic to edit the whole metadata

JuaniRios:
## What?
- change the `edit_metadata` extrinsic to be able to change all of the fields of `ProjectMetadata`

## Why?
- Issuers are likely to make mistakes when first creating the project

## How?
- if the project hasn't started the evaluation, the issuer can call edit_metadata with a copy of `ProjectMetadata` that contains optional fields. Each some field constitutes a replacement on the stored metadata. This new struct is called `OptionalProjectMetadata`.

## Testing?
`edit_metadata()`


vstam1 (changes requested):
Why have this completely new OptionalMetadata? Why not do the "initial metadata" struct over again for editing and just replace metadata instead of this difficult "updating". I think you are also missing the updates to project_details and bucket based on metadata, so create and edit could be quite similar. So:
- Validating metadata and updating bucket based on metadata and inserting metadata in storage  could be single function that is used by both create and edit.
- Create details in create, update details in edit.
- Create does all the transfers and deposits and updating project_id counter etc, while create only takes in Project_id

JuaniRios:
> Why have this completely new OptionalMetadata? Why not do the "initial metadata" struct over again for editing and just replace metadata instead of this difficult "updating". I think you are also missing the updates to project_details and bucket based on metadata, so create and edit could be quite similar. So:
> 
> * Validating metadata and updating bucket based on metadata and inserting metadata in storage  could be single function that is used by both create and edit.
> * Create details in create, update details in edit.
> * Create does all the transfers and deposits and updating project_id counter etc, while create only takes in Project_id

- I was thinking it from the PolkadotJS perspective where it would have been annoying to fill a whole project metadata with the same fields when you wanted to change just one, but I see your point. Probably the UI will do the heavy lifting for the user, so better just repeat the struct.

- Good point, I'll update projectDetails.
- I didn't get the last point

vstam1:
For the last point, a lot of the logic between do_create and edit_metadata will be similar if you just use ProjectMetadataOf as input. So you can probably move a lot of the logic to a seperate function call that function for both create and edit

pallets/funding/src/benchmarking.rs
```diff
@@ -458,24 +458,98 @@ mod benchmarks {
...
+						BalanceOf::<T>::try_from(10_000 * US_DOLLAR)
+							.unwrap_or_else(|_| panic!("Failed to create BalanceOf")),
+					),
+				),
+				phantom: Default::default(),
+			},
+			participation_currencies: vec![AcceptedFundingAsset::USDT, AcceptedFundingAsset::USDC].try_into().unwrap(),
+			funding_destination_account: issuer_funding.clone().clone(),
+			offchain_information_hash: Some(hashed(format!("{}-{}", METADATA, 69)).into()),
+		};
```
vstam1, line 538: I think we can again use the initial project metadata to benchmark this extrinsic? So we don't have to have this 80 line monster haha

JuaniRios: Initial one has a lot of None fields. This is benchmarking changing it to one that is completely different and bigger when encoded, so probably the proof size would change

PR #208, opened 2024-03-22
Feature/plmc 474 add extrinsic to remove project before evaluation started

JuaniRios:
## What?
- Adds an extrinsic to remove from storage a project before it started the evaluation

## Why?
- We don't want to bloat storage with projects that the issuer know he will never start. 

## How?
- if the project is not frozen, then remove it from storage

## Testing?
`remove_extrinsic()`

## Anything Else?
- Also wrote a benchmark for it


vstam1 (commented):
Should probably add a payment for creating a project as an issuer can now relatively freely create and delete projects?

vstam1:
I was thinking about an actual payment (not high but high enough to actually prevent creation + deletion and maybe also to prevent spamming of bullshit projects)

JuaniRios:
> I was thinking about an actual payment (not high but high enough to actually prevent creation + deletion and maybe also to prevent spamming of bullshit projects)

Only institutional credentials can create projects, that means an actual business KYB and not a person KYC. I think this is enough to prevent bullshit project.
And about the spam of creation/deletion, we already charge PLMC for the execution, so if they want to spam then better for us cause they pump up the transaction volume

PR #209, opened 2024-03-25
Feature/plmc 466 research if storage layers dont get dropped when calling

JuaniRios:
## What?
- Make sure our inner logic functions don't affect storage if they return an error

## Why?
- We call those functions directly in on_initialize, and so if they failed, the changes to storage would have been permanent

## How?
- add `#[transactional]` to all do_functions that returned a `DispatchResult`

## Testing?
`transactional_on_initialize()`

### DIFF
--- a/pallets/funding/src/functions.rs
+++ b/pallets/funding/src/functions.rs
@@ -18,12 +18,18 @@
 
 //! Functions for the Funding pallet.
 use crate::ProjectStatus::FundingSuccessful;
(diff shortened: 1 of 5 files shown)


vstam1 (approved):
Don't think we need the `#[transactional]` for settlement functions as they will be called directly via extrinsics.

JuaniRios:
> Don't think we need the `#[transactional]` for settlement functions as they will be called directly via extrinsics.

We don't know if in the future our automated system will call inner functions directly. And if whoever does it doesn't know about this transactional stuff, they can fuck up bad.
I don't see an issue on having everything transactional

lrazovic (approved):

PR #217, opened 2024-03-27
Feat/millis timestamp

lrazovic:
## What?

The timestamp pallet is storing the `now` as milliseconds, so we have to get the expiration from the JWT as milliseconds as well.

### DIFF
--- a/scripts/zombienet/native/local-testnet.toml
+++ b/scripts/zombienet/native/local-testnet.toml
@@ -4,7 +4,7 @@ provider = "native"
 
 # Using Rococo 1.7.0 as relay.
 [relaychain]
-default_command = "../polkadot-sdk/target/release/polkadot"
+default_command = "./bin/polkadot"
 chain = "rococo-local"
 
 [[relaychain.nodes]]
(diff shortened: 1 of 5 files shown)


JuaniRios (approved):
lgtm

scripts/zombienet/native/local-testnet.toml
```diff
@@ -2,31 +2,41 @@
 timeout = 1000
 provider = "native"
 
+# Using Rococo 1.7.0 as relay.
 [relaychain]
-default_command = "polkadot"
+default_command = "../polkadot-sdk/target/release/polkadot"
```
vstam1, line 7: Could we add a bin folder that is ignored by git where we can personally add these binaries?

vstam1 (approved):

scripts/zombienet/native/local-testnet.toml
lrazovic: Sure!

JuaniRios (approved):

scripts/zombienet/native/local-testnet.toml
lrazovic: Addressed and tested in https://github.com/Polimec/polimec-node/pull/217/commits/82f9d3f56dc25fd0654950924e97c6563cd4ea98. 

Should also the polimec-node point to the bin folder or should I leave it to` ./target/release`?

PR #218, opened 2024-03-27
feature/plmc-483-write-missing-unit-tests-on-pallet-funding-based-on-the

JuaniRios:
## What?
- Split round modules into separate files, and extrinsic/flow tests into separate modules to see more easily our code coverage.
    - This made some tests more verbose or duplicated, but I think the tradeoff is worth it for reviewing the code.
    - We might condense them sometime later.
- Add more tests for the application and evaluation rounds

## Why?
- We are starting the evaluation and releasing the whole pallet soon, so we need to test as much as possible

## Anything Else?
Only application and evaluation files have the new structure. The other ones will be structured once we review and write more tests on them in a separate PR

## How to review
Review the application and evaluation test files. The other ones are just copy pasta from the big tests.rs. file

See the following issue for a description of the logical paths we wanted to check:
https://linear.app/polimec/issue/PLMC-482/write-down-all-possible-logic-branches-to-test


pallets/funding/src/tests/1_application.rs
```diff
@@ -0,0 +1,1321 @@
...
+			let next_project_id = inst.execute(|| NextProjectId::<TestRuntime>::get());
+			assert_eq!(project_id, next_project_id - 1);
+			let projects_details = inst.execute(|| ProjectsDetails::<TestRuntime>::iter_keys().collect_vec());
+			let project_metadatas = inst.execute(|| ProjectsMetadata::<TestRuntime>::iter_keys().collect_vec());
+			assert_eq!(projects_details, vec![project_id]);
+			assert_eq!(project_metadatas, vec![project_id]);
+		}
+
+		#[test]
+		fn multiple_fields_edited() {
```
vstam1, line 707: Not sure if this test makes much sense as it is using the same logic as create_project for validation. Think just checking if setting the same metadata and some different metadata is possible. I don't think we care a lot about if we are able to change a single item, as we are not comparing the two different Metadata structures, right?

pallets/funding/src/tests/1_application.rs
```diff
@@ -0,0 +1,1321 @@
...
+			assert_eq!(stored_bucket, new_bucket);
+			// Event emitted
+			inst.execute(|| {
+				find_event!(TestRuntime, Event::<TestRuntime>::MetadataEdited{project_id, ref metadata}, project_id == 0, metadata == &new_metadata);
+			});
+		}
+	}
+
+	#[cfg(test)]
+	mod failure {
```
vstam1, line 849: We should probably remove all duplicate tests that check certain logic in the project_validation function as its the exact same logic being used in create_project. So only do tests here that are actually relevant to the lines that are not covered by create_project. So:
Issuer and project_frozen

vstam1: Or like Albert Einstein said: "Insanity is doing the same thing over and over again and expecting a different result" 😄

pallets/funding/src/tests/1_application.rs
```diff
@@ -0,0 +1,1321 @@
...
+				RuntimeOrigin::signed(ISSUER_1),
+				jwt.clone(),
+				project_id
+			)));
+			inst.execute(|| {
+				assert!(ProjectsDetails::<TestRuntime>::get(project_id).is_none());
+				assert!(ProjectsMetadata::<TestRuntime>::get(project_id).is_none());
+				assert!(Buckets::<TestRuntime>::get(project_id).is_none());
+				assert!(DidWithActiveProjects::<TestRuntime>::get(generate_did_from_account(ISSUER_1)).is_none());
+			});
```
vstam1, line 1194: Not sure if this makes sense, as we do not care about an off-chain hash in the removal process, right?

pallets/funding/src/tests/1_application.rs
```diff
@@ -0,0 +1,1321 @@
...
+			inst.execute(|| {
+				assert_noop!(
+					crate::Pallet::<TestRuntime>::remove_project(
+						RuntimeOrigin::signed(ISSUER_1),
+						jwt.clone(),
+						project_id
+					),
+					Error::<TestRuntime>::Frozen
+				);
+			});
```
vstam1, line 1208: Same logic as "evaluation_already_started" in the failed tests. We should use 1, I prefer the negative (so in failure)

vstam1 (commented):
Initial Review of everything except evaluation tests (and of course 3-8 tests). Will do evaluation tomorrow morning

pallets/funding/src/tests/1_application.rs
```diff
@@ -0,0 +1,1321 @@
...
+					one_currency_3,
+					two_currencies_1,
+					two_currencies_2,
+					two_currencies_3,
+					three_currencies,
+				];
+
+				let test_1 = with_different_metadata(one_currency_1);
+				let test_2 = with_different_metadata(one_currency_2);
+				assert!(test_1.offchain_information_hash != test_2.offchain_information_hash);
```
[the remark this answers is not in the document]
JuaniRios: This assert was there because `with_different_metadata` was not behaving as expected at first. But now that we removed the requirement of having unique hashes on each project we can remove all this

pallets/funding/src/tests/2_evaluation.rs
```diff
@@ -0,0 +1,552 @@
...
+			let evaluation_plmc = MockInstantiator::calculate_evaluation_plmc_spent(evaluations.clone());
+			let evaluation_existential = evaluation_plmc.accounts().existential_deposits();
+			inst.mint_plmc_to(evaluation_plmc);
+			inst.mint_plmc_to(evaluation_existential);
+
+			let project_id = inst.create_evaluating_project(project_metadata.clone(), ISSUER_2);
+			inst.evaluate_for_users(project_id, evaluations.clone()).unwrap();
+
+			let old_price = <TestRuntime as Config>::PriceProvider::get_price(PLMC_FOREIGN_ID).unwrap();
+			PRICE_MAP.with_borrow_mut(|map| map.insert(PLMC_FOREIGN_ID, old_price * 2.into()));
```
vstam1, line 71: Is it possible that the old price was already the initial price / 2 so now initial_price / 2 * 2 = initial_price?

pallets/funding/src/tests/1_application.rs
```diff
@@ -0,0 +1,1321 @@
...
+			fn multiple_creations_different_issuers() {
+				let mut inst = MockInstantiator::new(Some(RefCell::new(new_test_ext())));
+				let mut issuer = ISSUER_1;
+				for _ in 0..512 {
+					let project_metadata = default_project_metadata(inst.get_new_nonce(), issuer);
+					inst.create_evaluating_project(project_metadata, issuer);
+					inst.advance_time(1u64).unwrap();
+					issuer += 1;
+				}
+			}
```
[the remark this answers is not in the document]
JuaniRios: Yea I guess is a kind of fussing. We check if anything weird happens when we populate that storage map with a lot of projects. 
Not testing of anything in particular but more of a sanity check

pallets/funding/src/tests/1_application.rs
```diff
@@ -0,0 +1,1321 @@
...
+			for i in 1..=100 {
+				project_metadata.auction_round_allocation_percentage = Percent::from_percent(i);
+				issuer += 1;
+				let ed = MockInstantiator::get_ed();
+				inst.mint_plmc_to(vec![UserToPLMCBalance::new(issuer, ed * 2)]);
+				let jwt = get_mock_jwt(issuer, InvestorType::Institutional, generate_did_from_account(issuer));
+				assert_ok!(inst.execute(|| {
+					Pallet::<TestRuntime>::create_project(RuntimeOrigin::signed(issuer), jwt, project_metadata.clone())
+				}));
+			}
```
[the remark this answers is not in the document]
JuaniRios: Yea this tests would make more sense as a funding tests and not a creation test. Will delete

pallets/funding/src/tests/2_evaluation.rs
vstam1: Can we add a "set_price" function to the PriceProvider that does set the price in PRICE_MAP. Just for a cleaner approach.

pallets/funding/src/tests/1_application.rs
```diff
@@ -0,0 +1,1321 @@
...
+			let issuer: AccountId = ISSUER_1;
+			let mut counter: u8 = 0u8;
+			let mut with_different_hash = |mut project: ProjectMetadataOf<TestRuntime>| {
+				let mut binding = project.offchain_information_hash.unwrap();
+				let h256_bytes = binding.as_fixed_bytes_mut();
+				h256_bytes[0] = counter;
+				counter += 1u8;
+				project.offchain_information_hash = Some(binding);
+				project
+			};
```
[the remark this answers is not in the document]
JuaniRios: will remove this as well as the nonce on the default project metadata function since we can have duplicated hashes now

pallets/funding/src/tests/2_evaluation.rs
```diff
@@ -0,0 +1,552 @@
...
+			assert_eq!(project_status, ProjectStatus::EvaluationFailed);
+		}
+	}
+
+	#[cfg(test)]
+	mod failure {
+		use super::*;
+
+		#[test]
+		fn round_fails_after_not_enough_bonds() {
```
vstam1, line 86: Not sure if this function is necessary? We are already testing this case in a way in the test above. Or do you think its important for seperation of concerns.

pallets/funding/src/tests/1_application.rs
```diff
@@ -0,0 +1,1321 @@
...
+			// A project is "inactive" after the funding succeeds
+			inst.start_evaluation(3, ISSUER_1).unwrap();
+			inst.evaluate_for_users(3, default_evaluations()).unwrap();
+			inst.start_auction(3, ISSUER_1).unwrap();
+			inst.bid_for_users(3, default_bids()).unwrap();
+			inst.start_community_funding(3).unwrap();
+			inst.contribute_for_users(3, default_community_buys()).unwrap();
+			inst.start_remainder_or_end_funding(3).unwrap();
+			inst.contribute_for_users(3, default_remainder_buys()).unwrap();
+			inst.finish_funding(3).unwrap();
```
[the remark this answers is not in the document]
JuaniRios: Failing projects panic with our inst function, and I also want to do some `assert_noop`s in the middle of the flow.

But specifically for the last test of that function I could definitely replace it with the inst function

pallets/funding/src/tests/2_evaluation.rs
```diff
@@ -0,0 +1,552 @@
...
+#[cfg(test)]
+mod start_evaluation_extrinsic {
+	use super::*;
+
+	#[cfg(test)]
+	mod success {
+		use super::*;
+
+		#[test]
+		fn evaluation_starts() {
```
vstam1, line 138: Think it might be good to check all the phase_transition_points that are set. See if they are "correct" as its the biggest change the do_start_evaluation should make.

pallets/funding/src/tests/1_application.rs
JuaniRios: fair point, but might be good to decouple the desired behavior with the actual implementation no?

pallets/funding/src/tests/2_evaluation.rs
```diff
@@ -0,0 +1,552 @@
...
+				RuntimeOrigin::signed(issuer),
+				jwt,
+				project_id
+			)));
+			assert_eq!(inst.get_project_details(project_id).status, ProjectStatus::EvaluationRound);
+
+			inst.execute(|| {
+				assert_noop!(
+					PolimecFunding::start_evaluation(
+						RuntimeOrigin::signed(issuer + 1),
```
vstam1, line 252: Small nit, can we put here ISSUER_2 instead of issuer + 1?

pallets/funding/src/tests/2_evaluation.rs
```diff
@@ -0,0 +1,552 @@
...
+			let plmc_existential_deposits = necessary_plmc.accounts().existential_deposits();
+
+			inst.mint_plmc_to(necessary_plmc.clone());
+			inst.mint_plmc_to(plmc_existential_deposits);
+
+			inst.execute(|| {
+				frame_system::Account::<TestRuntime>::mutate(EVALUATOR_1, |account| {
+					account.data.frozen = necessary_plmc[0].plmc_amount;
+				});
+			});
```
vstam1, line 341: Would prefer to see a freeze using NativeCurrency (or via the implemented pallet directly) so we are actually checking against the correct interface. All other pallets use this pallet balances interface as well, so we do not know if something else would prevent us from using frozen tokens in pallet balances.

pallets/funding/src/tests/1_application.rs
JuaniRios: Like mentioned before, the reasoning here was decoupling the intended behavior from the tech implementation. If @lrazovic also thinks we should remove it and trust we will use the same implementation on both logic paths then let's do it

pallets/funding/src/tests/2_evaluation.rs
JuaniRios: We're creating a second project, evaluating on that, and THEN storing the old price, so we don't care about the initial_price no?

pallets/funding/src/tests/2_evaluation.rs
JuaniRios: This is a very critical component we need to test so I don't mind implicitly testing it twice if the trade-off is better readability

pallets/funding/src/tests/2_evaluation.rs
JuaniRios: Did it on purpose to make it clear we want another issuer, but sure

pallets/funding/src/tests/2_evaluation.rs
JuaniRios: Was inaccessible through the currency type on funding pallet, but I guess since I have access to the mock runtime I can do it from there

pallets/funding/src/tests/1_application.rs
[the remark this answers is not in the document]
JuaniRios: I'd like to keep this one

pallets/funding/src/tests/1_application.rs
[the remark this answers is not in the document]
JuaniRios: Ah wait, we can have failing projects with the `create_finished_project` function, good catch!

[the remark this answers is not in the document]
JuaniRios: Looks like I can't cleanly use the inst function. Ill keep it as is for readability's sake

pallets/funding/src/tests/1_application.rs
JuaniRios: CleanShot 2024-04-03 at 13 11 17@2x
That's already the case no?

pallets/funding/src/tests/2_evaluation.rs
vstam1: Its more that the second `let old_price` is the price that is set on the previous PRICE_MAP insert, so I was wondering if this was intended behaviour. So its currently the case that for example:
PLMC price at line 53 is 2
Then at line 56 its set to 1
Then PLMC price at line 68 is still 1.
And PLMC price is then set to 2 again at 71.

JuaniRios: Yea thats fine. 
What we care is :
Case 1) we evaluate at price x, round ends with price x/2
Case 2) we evaluate at price x, round ends with price x*2.

Remember that the evaluation is made in USD terms, which don't change when we change the price of PLMC, and we are minting the required PLMC to evaluate after each price change.

Makes sense?

pallets/funding/src/tests/mod.rs
```diff
@@ -0,0 +1,425 @@
...
+		CurrencyMetadata {
+			name: BoundedVec::try_from("Contribution Token TEST".as_bytes().to_vec()).unwrap(),
+			symbol: BoundedVec::try_from("CTEST".as_bytes().to_vec()).unwrap(),
+			decimals: ASSET_DECIMALS,
+		}
+	}
+	pub fn default_project_metadata(issuer: AccountId) -> ProjectMetadataOf<TestRuntime> {
+		let bounded_name = BoundedVec::try_from("Contribution Token TEST".as_bytes().to_vec()).unwrap();
+		let bounded_symbol = BoundedVec::try_from("CTEST".as_bytes().to_vec()).unwrap();
+		let metadata_hash = hashed(format!("{}", METADATA));
```
lrazovic, line 104:
```suggestion
        let metadata_hash = hashed(METADATA.to_string());  
```

pallets/funding/src/tests/7_settlement.rs
lrazovic, line 1: nit: in this file every` .clone()` in the `RuntimeOrigin::signed(...)` should not be needed

pallets/funding/src/tests/1_application.rs
```diff
@@ -0,0 +1,1148 @@
...
+						generate_did_from_account(ISSUER_1),
+					),
+					Error::<TestRuntime>::BadMetadata(MetadataError::AllocationSizeError)
+				);
+			});
+		}
+
+		#[test]
+		fn invalid_ticket_sizes() {
+			let mut correct_project = default_project_metadata(ISSUER_1);
```
lrazovic, line 364:
```suggestion
			let correct_project = default_project_metadata(ISSUER_1);
```

pallets/funding/src/tests/3_auction.rs
```diff
@@ -0,0 +1,1299 @@
...
+}
+
+#[test]
+fn bid_with_asset_not_accepted() {
+	let mut inst = MockInstantiator::new(Some(RefCell::new(new_test_ext())));
+	let project_id =
+		inst.create_auctioning_project(default_project_metadata(ISSUER_1), ISSUER_1, default_evaluations());
+	let bids = vec![BidParams::<TestRuntime>::new(BIDDER_1, 10_000, 1u8, AcceptedFundingAsset::USDC)];
+
+	let did = generate_did_from_account(bids[0].bidder.clone());
```
lrazovic, line 655:
```suggestion
	let did = generate_did_from_account(bids[0].bidder);
```

lrazovic (approved):
The refactored tests look great! Just a few minor things to address

pallets/funding/src/tests/1_application.rs
lrazovic: I would be on the side of testing the "tech implementation" here and the "intended behavior" in the various integration tests via xcm-emulator and/or chopsticks

pallets/funding/src/tests/1_application.rs
```diff
@@ -0,0 +1,1148 @@
...
+			let projects_details = inst.execute(|| ProjectsDetails::<TestRuntime>::iter_keys().collect_vec());
+			let project_metadatas = inst.execute(|| ProjectsMetadata::<TestRuntime>::iter_keys().collect_vec());
+			assert_eq!(projects_details, vec![project_id]);
+			assert_eq!(project_metadatas, vec![project_id]);
+		}
+
+		#[test]
+		fn multiple_fields_edited() {
+			let mut inst = MockInstantiator::new(Some(RefCell::new(new_test_ext())));
+			let project_metadata = ProjectMetadata {
```
vstam1, line 566: Can we make this from default

vstam1 (approved):
Approved after:
- [x] fixing last open comments from me/Leo
- [x] fix all warnings etc 
- [x] fmt

I would like to see some comments on the top of the tests (maybe only the more complicated once)

PR #219, opened 2024-03-28
add plmc price to oracle

vstam1:
## What?
- Added XT and MexC fetchers to the Oracle to fetch the PLMC price
- Added testing of fetchers against real time apis, so that we are quickly updated if some api changes breaks our oracle impl.

## Why?
We need the PLMC price for all our funding calculations.

## Testing?
Tests are updated to account of new fetchers + updated tests to check against real apis


JuaniRios (approved):
lgtm

PR #226, opened 2024-04-09
Feature: Dispenser Pallet

vstam1:
## What?
This PR introduces the dispenser-pallet, that can be used to get locked PLMC tokens for account with a valid credential. 

## Why?
To enable initial user growth on the Polimec network, we allow a certain amount of users to dispense tokens.

## How?
The pallet has a pallet account that can be funded to distribute tokens. The token amount that can be dispensed is specified in the `DispenseAmount` storage item and can be set by an admin account using the `set_dispense_amount` extrinsic. The initial amount is determined by the InitialDispenseAmount. 

A user can call the `dispense` extrinsic to dispense tokens. There are a couple of checks to see if the user is able to dispense. 
First, we check if the user has a valid credential and then we check if the Did in the credential has already dispensed tokens. We also check if the funding wallet still has tokens to distribute. 

#### How do Signed Extensions work? 
Every extrinsic that is submitted to a node, is first validated before it is added to the transaction pool and broadcasted to other nodes. The validation logic is written in Signed Extensions. If an extrinsic is valid, is being broadcasted to all other nodes. The validity of extrinsics is being tracked while it's in the transaction pool (it can get invalid by passing its mortality time for example). Once a collator starts building a block with this extrinsic, it first executes the pre_dispatch check, which should have similar logic as the validation check. 

#### How to use SignedExtensions for free execution (in theory)?
Parity introduced the new `SkipCheckIfFeeless` signed extrinsic which is a wrapper around other signed extensions. A call can be marked as potentially free using `feeless_if` (see dispense extrinsic). The `feeless_if` implementation has some logic attached to see if the call is actually free. If `feeless_if` returns `true`, `SkipCheckIfFeeless` skips the inner signed extension check. This can be used to skip the `ChargeTransactionPayment` extension, which is responsible for charging fees. So in Theory, just wrapping this `ChargeTransactionPayment` extension with `SkipCheckIfFeeless` and correctly implementing `feeless_if`, should make the call free.

#### Free execution in practice?
`SkipCheckIfFeeless` doesn't work out of the box. There were two problems we ran into:

- SkipCheckIfFeeless did not implement the validate function, which would skip the validation logic for fee payment for every transaction. So transactions were only checked for fee payment in the pre_dispatch function, which opens up a transaction pool DoS attack vector. We fixed this by implementing this Signed Extension ourselves with the validate function. Parity also implements it now: https://github.com/paritytech/polkadot-sdk/pull/3993.

- The CheckNonce extension has a check that the user paid for the Nonce storage by checking if the accounts providers and sufficients are non-zero. For new accounts these are zero, so we had to skip this check for the dispense call. So I implemented the CheckNonce signed extension myself, with this extra logic. Parity acknowledged this problem and is discussing a general solution at https://github.com/paritytech/polkadot-sdk/issues/3991. I think we are fine with our current implementation, as users actually receive tokens that will be locked/vested over a period of 4 years and the accounts able to get the tokens are limited by the credentials. 

## Testing?
Normal unit tests + a test to check the Signed extensions in the integration tests.

## Anything Else?
Todo:
- [ ] Change the Admin account?


pallets/dispenser/src/tests.rs
```diff
@@ -0,0 +1,122 @@
...
+		ExtBuilder::default().build().execute_with(|| {
+			let jwt = get_mock_jwt(1, InvestorType::Retail, generate_did_from_account(1));
+			assert_ok!(Dispenser::dispense(RuntimeOrigin::signed(1), jwt.clone()));
+			assert_noop!(Dispenser::dispense(RuntimeOrigin::signed(1), jwt), Error::<Test>::DispensedAlreadyToDid);
+		});
+	}
+
+	#[test]
+	fn x_users_dispense_until_dispenser_is_empty() {
+		let x = 10;
```
JuaniRios, line 96: This test is failing

runtimes/politest/src/lib.rs
```diff
@@ -835,9 +838,11 @@ where
 			frame_system::CheckTxVersion::<Runtime>::new(),
 			frame_system::CheckGenesis::<Runtime>::new(),
 			frame_system::CheckEra::<Runtime>::from(generic::Era::mortal(period, current_block)),
-			frame_system::CheckNonce::<Runtime>::from(nonce),
+			pallet_dispenser::extensions::CheckNonce::<Runtime>::from(nonce),
```
lrazovic, line 841: Is this a temporary measure needed until https://github.com/paritytech/polkadot-sdk/issues/3991 gets fixed? In that case maybe let's add a `TODO` so we remember to go back to `frame_system::CheckNonce::<Runtime>::from(nonce)` as soon it's possible

pallets/dispenser/src/lib.rs
```diff
@@ -0,0 +1,187 @@
...
+				T::InvestorOrigin::ensure_origin(origin, &jwt, T::VerifierPublicKey::get())?;
+			ensure!(Dispensed::<T>::get(&did).is_none(), Error::<T>::DispensedAlreadyToDid);
+
+			let amount = DispenseAmount::<T>::get();
+			ensure!(CurrencyOf::<T>::free_balance(&Self::dispense_account()) >= amount, Error::<T>::DispenserDepleted);
+
+			let current_block = <frame_system::Pallet<T>>::block_number();
+			let length_as_balance = T::BlockNumberToBalance::convert(T::VestPeriod::get());
+			let locked_amount = amount.saturating_sub(T::FreeDispenseAmount::get());
+			let per_block = locked_amount / length_as_balance.max(sp_runtime::traits::One::one());
```
lrazovic, line 148: Is it safe to use the `/` here? Or should we use safe/checked math? What happens when `locked_amount / length_as_balance` is a float?

pallets/dispenser/src/tests.rs
```diff
@@ -0,0 +1,122 @@
...
+				Error::<Test>::DispenseAmountTooLow
+			);
+		});
+	}
+}
+
+mod dispense {
+	use super::*;
+	#[test]
+	fn user_can_dispense_for_free() {
```
JuaniRios, line 64: The mock runtime doesn't charge a fee for executing extrinsics. I assume this is to make pallet tests easier since a lot of pallets modify your balance outside of the fees charged.
You can double-check this with this simple test:
```rust
	#[test]
	fn sandbox() {
		ExtBuilder::default().build().execute_with(|| {
			assert_eq!(Balances::free_balance(1), 0);
			assert_ok!(System::remark(RuntimeOrigin::signed(1), vec![69, 69]));
		});
	}

```

I think you need to modify the mock runtime to actually charge fees (if that is even possible)

pallets/dispenser/src/lib.rs
```diff
@@ -0,0 +1,187 @@
...
+		PalletId,
+	};
+	use frame_system::pallet_prelude::*;
+	use sp_runtime::{traits::AccountIdConversion, Saturating};
+
+	/// Configure the pallet by specifying the parameters and types on which it depends.
+	#[pallet::config]
+	pub trait Config: frame_system::Config {
+		/// The Origin that has admin access to change the dispense amount.
+		type AdminOrigin: EnsureOrigin<Self::RuntimeOrigin>;
```
lrazovic, line 60: Nit: can we use the 

```
/// Comment
type
<White line>
```

syntax?

JuaniRios (changes requested):
Most important thing to test out is that the extrinsic can actually be called for free, and we are not really testing that in the pallet test because all extrinsics are free.

After the pallet test actually tests that, I suggest testing it by running a local node with zombienet.

lrazovic (changes requested):
Love to see the `feeless_if` + JWT check in action!

The logic looks good, just few comments here and there that I would like to discuss before merging this into `main`.

Also, the `try-runtime-cli` is complaining about the Funding Pallet: `Funding: On chain storage version StorageVersion(1) is set to non zero, while the pallet is missing the #[pallet::storage_version(VERSION)] attribute.` So maybe we should rebase this PR to include https://github.com/Polimec/polimec-node/pull/223

pallets/dispenser/src/lib.rs
```diff
@@ -0,0 +1,187 @@
...
+
+			let amount = DispenseAmount::<T>::get();
+			ensure!(CurrencyOf::<T>::free_balance(&Self::dispense_account()) >= amount, Error::<T>::DispenserDepleted);
+
+			let current_block = <frame_system::Pallet<T>>::block_number();
+			let length_as_balance = T::BlockNumberToBalance::convert(T::VestPeriod::get());
+			let locked_amount = amount.saturating_sub(T::FreeDispenseAmount::get());
+			let per_block = locked_amount / length_as_balance.max(sp_runtime::traits::One::one());
+
+			T::VestingSchedule::can_add_vesting_schedule(
```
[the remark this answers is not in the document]
vstam1: Its described in the vesting pallet implementation that it should always be called before the add_vesting_schedule. But it might be that its older code that was written before the transactionality of extrinsics, where the add_vesting_schedule should succeed after the transfer of tokens. I could probably remove it if you want?

pallets/dispenser/src/tests.rs
vstam1: Why exactly? We are testing the correct execution of the logic within the extrinsics. For the "free execution" tests, we have an integration test that checks the validate and pre_dispatch execution. And I ran the extrinsic on zombienet to see if execution was free, and this worked. So I thing these three together should show the correct working of the pallet.

pallets/dispenser/src/lib.rs
```diff
@@ -0,0 +1,187 @@
...
+
+#[cfg(test)]
+mod tests;
+
+pub mod weights;
+
+#[cfg(feature = "runtime-benchmarks")]
+mod benchmarking;
+
+pub type BalanceOf<T> = <CurrencyOf<T> as Currency<AccountIdOf<T>>>::Balance;
```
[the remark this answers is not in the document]
vstam1: We ware unfortunately stuck with Currency because of pallet vesting. We could probably implement both (currency for vesting and fungible for the transfer of tokens) but I rather keep it to 1 for simplicity. We can of course update once pallet_vesting is updated.

pallets/dispenser/src/lib.rs
vstam1: I think this is safe to use as I copied the logic from a parity pallet, but I can change it to safe math just to be sure!

lrazovic (approved):

pallets/dispenser/src/lib.rs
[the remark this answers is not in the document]
JuaniRios: My only worry is that we already encountered the possibility of executing logic in a non-transactional way, and that was running functions that touched storage inside on_initialize.

I think if we are sure this will always run inside an extrinsic which has transacitonal by default, then we can be pretty sure it will roll back on error and remove the can_add_vesting_schedule. But I cannot tell you 100% it will be transactional.

I leave it up to you if you want to remove it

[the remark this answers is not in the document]
vstam1: Lets keep it for now. Might do some double checks but its fine. Pretty sure this logic will mainly stay in an extrinsic as it is part of an extrinsic.

pallets/dispenser/src/tests.rs
JuaniRios: the test is called `user_can_dispense_for_free` but you are not testing that because there are no fees in this mock runtime, so if you think this is tested in other places like integration-tests, then I would remove this test and maybe leave a comment where this test can be found.

I see though that we have 2 dispenser tests called `dispenser_signed_extensions_pass_for_new_account` and `dispenser_works_with_runtime_values` But I don't see it testing that a user with 0 balance can call dispense, while other calls would require fees/ED.

JuaniRios (approved):

PR #229, opened 2024-04-11
✅ Comprehensive community round tests (518)

JuaniRios:
## What?
- Found and fixed critical bug in the price calculation
- Wrote and modified tests to cover more paths and make code clearer

## How?
- PLMC and Funding asset returns only happen when they are higher than ED. Also wap is taken as min_price when wap is lower than min_price. This can happen due to rounding errors when bucket is increased only by rejected bids.

## Testing?
`cargo test --features runtime-benchmarks`

## Anything else?
check https://linear.app/polimec/issue/PLMC-502/community for documentation on all paths covered


JuaniRios:
https://linear.app/polimec/issue/PLMC-518/comprehensive-community-round-tests

pallets/funding/src/tests/4_community.rs
```diff
@@ -1,1224 +1,1394 @@
...
+					log::debug!("created");
+					assert_ok!(contribute(&mut inst, project_id, 1));
+				});
+
+				let project_id = create_project(&mut inst);
+				log::debug!("created");
+				previous_projects_created = projects_participated_amount;
+
+				// 0x multiplier should fail
+				// Professional bids: 0x multiplier should fail
```
vstam1, line 699: Retail

pallets/funding/src/tests/4_community.rs
```diff
@@ -1,1224 +1,1394 @@
...
+				)
+			};
+			let contribute = |inst: &mut MockInstantiator, project_id, multiplier| {
+				let jwt = get_mock_jwt(BUYER_1, InvestorType::Retail, generate_did_from_account(BUYER_1));
+				let wap = inst.get_project_details(project_id).weighted_average_price.unwrap();
+				let contributor_plmc = MockInstantiator::calculate_contributed_plmc_spent(
+					vec![(BUYER_1, 1_000 * ASSET_UNIT, Multiplier::force_new(multiplier)).into()],
+					wap,
+				);
+				let bidder_usdt = MockInstantiator::calculate_contributed_funding_asset_spent(
```
vstam1, line 658: contributor_usdt?

pallets/funding/src/tests/4_community.rs
```diff
@@ -1,1224 +1,1394 @@
...
+					assert_err!(
+						contribute(&mut inst, project_id, multiplier),
+						Error::<TestRuntime>::ForbiddenMultiplier
+					);
+				}
+			}
+		}
+
+		#[test]
+		fn did_with_losing_bid_can_contribute() {
```
vstam1, line 732: We maybe should skip the check on bidder 4 and 6 that are evaluators, as we do not have logic that would prevent an evaluator from contributing, right? Only people with failing bids? Something that you could test is: Evaluator does bid and uses eval PLMC to bid. Bid fails so eval plmc should be returned and can be used for contribution. Might also be an edge case, where Evaluation bond used for failing Bid should not be returned to user as unbonded PLMC? Currently I could evaluate and then create failing bid and get all my unslashable PLMC back before project ends.

pallets/funding/src/tests/4_community.rs
```diff
@@ -1,1224 +1,1394 @@
...
+			bid_should_fail(BIDDER_3, InvestorType::Retail, BIDDER_3);
+			bid_should_fail(BUYER_3, InvestorType::Retail, BIDDER_3);
+		}
 
-		log::debug!("creating {} new projects", projects_participated_amount - previous_projects_created);
+		#[test]
+		fn per_credential_type_ticket_size_minimums() {
+			let mut inst = MockInstantiator::new(Some(RefCell::new(new_test_ext())));
+			let mut project_metadata = default_project_metadata(ISSUER_1);
+			project_metadata.bidding_ticket_sizes = BiddingTicketSizes {
```
vstam1, line 1028: Can skip bidding ticket size here, as we are testing community round?

pallets/funding/src/tests/4_community.rs
```diff
@@ -1,1224 +1,1394 @@
...
+			});
+		}
 
-		let project_id = create_project(&mut inst);
-		log::debug!("created");
-		previous_projects_created = projects_participated_amount;
+		#[test]
+		fn per_credential_type_ticket_size_maximums() {
+			let mut inst = MockInstantiator::new(Some(RefCell::new(new_test_ext())));
+			let project_metadata = ProjectMetadata {
```
vstam1, line 1110: From default

pallets/funding/src/tests/4_community.rs
```diff
@@ -1,1224 +1,1394 @@
...
-		});
 
-		// Multipliers that should work
-		for multiplier in 1..=max_allowed_multiplier {
-			log::debug!("success? - multiplier: {}", multiplier);
-			assert_ok!(contribute(&mut inst, project_id, multiplier));
+			let jwt = get_mock_jwt(BUYER_1, InvestorType::Retail, generate_did_from_account(BUYER_1));
+			let contribution = ContributionParams::new(BUYER_1, 1_000 * ASSET_UNIT, 1u8, AcceptedFundingAsset::USDT);
+			let wap = inst.get_project_details(project_id).weighted_average_price.unwrap();
+			dbg!(&wap);
```
vstam1, line 1287: rm dbg

pallets/funding/src/tests/4_community.rs
```diff
@@ -1,1224 +1,1394 @@
...
+
+			// 1 unit less native asset than needed
+			let plmc_funding = MockInstantiator::calculate_contributed_plmc_spent(vec![contribution.clone()], wap);
+			let plmc_existential_deposits = plmc_funding.accounts().existential_deposits();
+			inst.mint_plmc_to(plmc_funding.clone());
+			inst.mint_plmc_to(plmc_existential_deposits.clone());
+			inst.execute(|| Balances::burn_from(&BUYER_1, 1, Precision::BestEffort, Fortitude::Force)).unwrap();
```
vstam1, line 1294: Instead of burning, maybe subtract 1 from mint.

vstam1 (commented):
Good first changes! Did an review on tests, but will do another on maybe missing paths.

pallets/funding/src/functions.rs
```diff
@@ -2025,14 +1983,18 @@ impl<T: Config> Pallet<T> {
...
 		} else {
 			accepted_bids
 				.iter()
 				.map(calc_weighted_price_fn)
 				.fold(Zero::zero(), |a: T::Price, b: T::Price| a.saturating_add(b))
 		};
+		// If the first bucket was sold out with rejected bids, the wap might be slightly lower than min_price due to rounding.
+		if weighted_token_price < project_metadata.minimum_price {
+			weighted_token_price = project_metadata.minimum_price;
+		}
```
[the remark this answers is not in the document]
JuaniRios: Maybe my comment wasn't clear.
What I mean is that if all the first bucket is with accepted bids, and the second bucket is with post-random-end bids, then we will try to calculate the wap from the weights and this introduces the rounding error. Let me fix the comment to make it more understandable

pallets/funding/src/functions.rs
```diff
@@ -2075,12 +2042,16 @@ impl<T: Config> Pallet<T> {
...
+				let plmc_bond_returned = bid.plmc_bond.saturating_sub(plmc_bond_needed);
+				// If the free balance of a user is zero and we want to send him less than ED, it will fail.
+				if plmc_bond_returned > T::ExistentialDeposit::get() {
+					T::NativeCurrency::release(
+						&HoldReason::Participation(project_id).into(),
+						&bid.bidder,
+						plmc_bond_returned,
+						Precision::Exact,
+					)?;
+				}
```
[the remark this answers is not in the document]
JuaniRios: This is the logic for placing a hold. It is called with `Protect` and `Force`.
The case for erroring out when there's no ED left after hold is when there is only 1 provider. And if I understand correctly, we could in the future introduce pallets that provide for the existence of the account, making it possible for it to live without an ED.

Therefore I prefer to be safe and do a check before releasing so we don't halt the whole project. I don't think most people will be getting a return below ED anyway, and if they do they shouldn't mind it not being returned.

@lrazovic any thoughts?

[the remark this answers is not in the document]
JuaniRios: For reference here is the logic for releasing an asset. It always checks the amount free + the released amount will be bigger than ED.

pallets/funding/src/tests/4_community.rs
```diff
@@ -1,1224 +1,1394 @@
...
+					project_id,
+					usable_ct / 2,
+					1u8.try_into().unwrap(),
+					AcceptedFundingAsset::USDT,
+				));
+			});
+
+			// If we mint what we just used, and mint what we cannot use due to being a slash deposit, \
+			// then we can use make a contribution that uses the same amount that we originally bonded on the evaluation
+			let spent_usable_plmc = (BOB, bobs_usable_bond / 2).into();
```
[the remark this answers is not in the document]
JuaniRios: I wanted to test using partial and full amount of usable evaluation-bonded PLMC.

In this test we:
1) Try using all of it for a contribution - fails because we need to leave the slash deposit
2) Try partially using half of it - Succeeds
3) Try minting the slashable deposit plus what we just spent so the test 1) can now pass.

I agree point 3 might be confusing, so maybe the 3rd test could be a different account doing the same evaluation, and then using all of the usable deposit without any extra mints?
And then a test 4) that tries to repeat it and fails?

pallets/funding/src/tests/4_community.rs
JuaniRios: Fair point, was discussed with the team and a ticket added for further post-mvp discussion:
https://linear.app/polimec/issue/PLMC-527/future-discussion-evaluation-bond-early-release

Main takeaway is that early release is not a problem.

Checking that evaluator-bidders can participate if the bid failed is IMO still a good edge case even if like you mention we do the check only on the bid participation and not the evaluation.

The test for using evaluation plmc after having a failed bid will be added in `evaluation_bond_counts_towards_contribution`.

pallets/funding/src/tests/4_community.rs
```diff
@@ -226,26 +233,82 @@ mod community_contribute_extrinsic {
...
+					RuntimeOrigin::signed(CARL),
+					get_mock_jwt(CARL, InvestorType::Retail, generate_did_from_account(CARL)),
+					project_id,
+					usable_ct,
+					1u8.try_into().unwrap(),
+					AcceptedFundingAsset::USDT,
+				));
+			});
+
+			// An evaluator that did a bid but it was not accepted at the end of the auction, can use that PLMC for contributing
```
vstam1, line 250: Would be good to move the logic after this comment to a separate test. Just for clarity sake.

pallets/funding/src/tests/4_community.rs
```diff
@@ -1306,6 +1249,10 @@ mod community_contribute_extrinsic {
 				inst.contribute_for_users(project_id_usdt, vec![dot_contribution.clone()]),
 				Error::<TestRuntime>::FundingAssetNotAccepted
 			);
+		}
+
+		#[test]
+		fn sandbox() {
```
vstam1: Remove sendbox

pallets/funding/src/tests/misc.rs
```diff
@@ -563,4 +563,39 @@ mod bug_hunting {
 		let new_project_details = inst.get_project_details(project_id);
 		assert_eq!(old_project_details, new_project_details);
 	}
+
+	// POC showing why the price calculation can be wrong when correct calculation would require more decimals than available
```
vstam1, line 567: Think we can remove this test.

vstam1 (approved):
Looks good for a couple of last things

pallets/funding/src/functions.rs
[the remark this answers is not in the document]
JuaniRios: If it fails due to ED, we can cover it. If it fails for some other reason we probably don't want to continue the project transition because it's unexpected behavior and we should fix it and then force run the community start.

So I think we can resolve as is

PR #235, opened 2024-04-15
🐛 Staking benchmarks

JuaniRios:
## What?
- Fix the staking benchmarks

## Why?
- To generate the weights with our own runtime. They only worked with mock

## How?
- Stopped using hardcoded values and assuming ED

## Testing?
`just dry-run-benchmarks`


pallets/funding/src/tests/mod.rs
lrazovic, line 1: Can you please remove these leftover comments?

lrazovic (approved):

PR #243, opened 2024-04-17
🐛 Fix benchmarks

JuaniRios:
## What?
- fix all benchmarks that errored out when running `just dry-run-benchmarks`

## How?
- Pallet funding required funding the ct treasury account with ed on the chain spec
- Linear release required using config types in the benchmark code instead of hardcoded values

## Testing?
- Run `just dry-run-benchmarks` and see if no errors pops up

### DIFF
--- a/nodes/parachain/src/chain_spec/politest.rs
+++ b/nodes/parachain/src/chain_spec/politest.rs
@@ -19,6 +19,7 @@
 //! Polimec Testnet chain specification
 
 use cumulus_primitives_core::ParaId;
+use frame_benchmarking::__private::traits::fungible::Inspect;
 use politest_runtime::{
 	pallet_parachain_staking::{
 		inflation::{perbill_annual_to_perbill_round, BLOCKS_PER_YEAR},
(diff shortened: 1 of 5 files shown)


nodes/parachain/src/chain_spec/politest.rs
```diff
@@ -19,6 +19,7 @@
 //! Polimec Testnet chain specification
 
 use cumulus_primitives_core::ParaId;
+use frame_benchmarking::__private::traits::fungible::Inspect;
```
lrazovic, line 22: Why `Inspect` is coming from `frame_benchmarking` and not from `frame_support::traits::tokens::fungible::Inspect`?

vstam1 (approved):

lrazovic (approved):
LGTM! Approved with one minor question

nodes/parachain/src/chain_spec/politest.rs
JuaniRios: oops, auto import crapped out

PR #244, opened 2024-04-17
🐛Fix overweight contribution

JuaniRios:
Built on top of #243 

## What?
- The proof sizes on ProjectsToUpdate for contribution_ends_round extrinsic seems way to high. We fixed them

## Why?
- We don't know for sure

## How?
- I changed from a BoundedVec to a single tuple for ProjectsToUpdate, that seemed to fix it, but when I reran the benchmark on main it produced a similar result to the tuple change, so the original error might have been a mistake on the machine.

## Anything Else?
- Even though the change of this PR probably wasn't the cause of the fix, I think it's worth to include since the boundedVec we use is always of length 1


test_weight.rs
lrazovic, line 1: Is this file needed?

pallets/funding/src/weights.rs
```diff
@@ -806,43 +815,52 @@ impl WeightInfo for () {
...
-	/// Storage: `StatemintAssets::Account` (r:2 w:2)
-	/// Proof: `StatemintAssets::Account` (`max_values`: None, `max_size`: Some(134), added: 2609, mode: `MaxEncodedLen`)
-	/// Storage: `PolimecFunding::ProjectsToUpdate` (r:9833 w:1)
-	/// Proof: `PolimecFunding::ProjectsToUpdate` (`max_values`: None, `max_size`: Some(27), added: 2502, mode: `MaxEncodedLen`)
-	/// The range of component `x` is `[1, 255]`.
+	/// Storage: `ForeignAssets::Asset` (r:1 w:1)
+	/// Proof: `ForeignAssets::Asset` (`max_values`: None, `max_size`: Some(210), added: 2685, mode: `MaxEncodedLen`)
+	/// Storage: `ForeignAssets::Account` (r:2 w:2)
+	/// Proof: `ForeignAssets::Account` (`max_values`: None, `max_size`: Some(134), added: 2609, mode: `MaxEncodedLen`)
+	/// Storage: `Funding::ProjectsToUpdate` (r:100 w:1)
```
lrazovic, line 844: 100 Reads are still a very high number, for sure we can optimize this. Let's create an Issue on Linear.

lrazovic (approved):

pallets/funding/src/weights.rs
JuaniRios: its 100 reads because we set the max update attempts to 100. We cannot optimize that

lrazovic:
Also, let's fix the warnings before merging:

```
warning: variable does not need to be mutable
   --> pallets/funding/src/benchmarking.rs:571:7
    |
571 |         let mut block_number: BlockNumberFor<T> = inst.current_block() + T::EvaluationDuration::get() + One::one();
    |             ----^^^^^^^^^^^^
    |             |
    |             help: remove this `mut`
    |
    = note: `#[warn(unused_mut)]` on by default

warning: for loop over an `Option`. This is more readably written as an `if let` statement
    --> pallets/funding/src/lib.rs:1246:37
     |
1246 |             for (project_id, update_type) in ProjectsToUpdate::<T>::take(now) {
     |                                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
     |
     = note: `#[warn(for_loops_over_fallibles)]` on by default
help: to check pattern in a loop use `while let`
     |
1246 |             while let Some((project_id, update_type)) = ProjectsToUpdate::<T>::take(now) {
     |             ~~~~~~~~~~~~~~~                         ~~~
help: consider using `if let` to clear intent
     |
1246 |             if let Some((project_id, update_type)) = ProjectsToUpdate::<T>::take(now) {
     |             ~~~~~~~~~~~~                         ~~~
     ```

PR #248, opened 2024-04-18
🐛 Fix pallet-xcm benchmarks

JuaniRios:
## What?
- Fix pallet-xcm benchmarks

## Why?
- So we can have accurate weights for our runtimes 

## How?
- Add some janky patches to go around the shitty code of the pallet.
- Most of that shitty code is fixed in the latest version of pallet-xcm, but we cannot use that now till we upgrade substrate version

## Testing?
`just dry-run-benchmarks`


JuaniRios:
* **#248** <a href="https://app.graphite.dev/github/pr/Polimec/polimec-node/248?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

Cargo.toml
```diff
@@ -237,4 +237,4 @@ polimec-runtime = { path = "runtimes/polimec" }
 asset-hub-polkadot-runtime = { git = 'https://github.com/polkadot-fellows/runtimes', default-features = false, rev = '59172d5fb8feea2646490618bead3b2b4fb06220' }
 polkadot-runtime-constants = { git = 'https://github.com/polkadot-fellows/runtimes', default-features = false, rev = '59172d5fb8feea2646490618bead3b2b4fb06220' }
 system-parachains-constants = { git = 'https://github.com/polkadot-fellows/runtimes', default-features = false, rev = '59172d5fb8feea2646490618bead3b2b4fb06220' }
-polkadot-runtime = { git = 'https://github.com/polkadot-fellows/runtimes', default-features = false, rev = '59172d5fb8feea2646490618bead3b2b4fb06220' }
+polkadot-runtime = { git = 'https://github.com/polkadot-fellows/runtimes', default-features = false, rev = '59172d5fb8feea2646490618bead3b2b4fb06220' }
```
vstam1, line 240: Nit: Remove this new line

runtimes/polimec/src/lib.rs
```diff
@@ -429,9 +429,35 @@ impl cumulus_pallet_aura_ext::Config for Runtime {}
 parameter_types! {
 	pub const RelayOrigin: AggregateMessageOrigin = AggregateMessageOrigin::Parent;
 }
+pub struct MockedChannelInfo;
```
vstam1, line 432: Can you add a todo comment to this struct why we need it and when we can remove it?
Might also be good to make a linear issue for all TODOs we need to do for next version upgrade. Haven't done it myself but just though of it.

runtimes/polimec/src/lib.rs
```diff
@@ -1195,17 +1221,27 @@ impl_runtime_apis! {
 			use cumulus_pallet_session_benchmarking::Pallet as SessionBench;
 			use xcm::latest::prelude::*;
 			use pallet_xcm::benchmarking::Pallet as PalletXcmExtrinsiscsBenchmark;
-			/// TODO: Update these benchmarks once we enable PLMC Teleportation
+			/// TODO: Update these janky implementations once we upgrade pallet_xcm. The jankyness is due to badly written pallet-xcm code
```
vstam1, line 1224: Can you keep the old TODO as well? And maybe remove the comment on badly written as its an open source repo 🤣

runtimes/politest/src/lib.rs
```diff
@@ -456,8 +456,36 @@ parameter_types! {
 	pub const RelayOrigin: AggregateMessageOrigin = AggregateMessageOrigin::Parent;
 }
 
+pub struct MockedChannelInfo;
```
vstam1, line 459: Same as other comment

runtimes/politest/src/lib.rs
```diff
@@ -1401,17 +1429,27 @@ impl_runtime_apis! {
 			use cumulus_pallet_session_benchmarking::Pallet as SessionBench;
 			use xcm::latest::prelude::*;
 			use pallet_xcm::benchmarking::Pallet as PalletXcmExtrinsiscsBenchmark;
-			/// TODO: Update these benchmarks once we enable PLMC Teleportation
+			/// TODO: Update these janky implementations once we upgrade pallet_xcm. The jankyness is due to badly written pallet-xcm code
```
vstam1, line 1432: Same

runtimes/polimec/src/lib.rs
```diff
@@ -1195,17 +1221,27 @@ impl_runtime_apis! {
...
+						RuntimeOrigin::root(),
+						Box::new(crate::xcm_config::AssetHubLocation::get()),
+						xcm::prelude::XCM_VERSION
+					).unwrap();
 					Some((
 						MultiAsset {
-							fun: Fungible(EXISTENTIAL_DEPOSIT),
-							id: Concrete(Parent.into())
+							fun: Fungible(ExistentialDeposit::get()),
+							id: Concrete(Here.into())
```
vstam1, line 1244: Why should this be Here? As Here describes PLMC in this context, right? Instead of Dot?

vstam1 (changes requested):
Nice that you got the pallet-xcm benchmarks to work. If I understand it correctly, the two problems we run into are:
- XCM version is not set? Is this fixed in future versions, and how?
- The channel info is normally not properly mocked? How do other chains fix this, as this should be a general problem right?

runtimes/polimec/src/lib.rs
JuaniRios: Benchmarks use Balances pallet to fund the accounts before the reserve transfer, so we only have the choice of specifying our native asset which was one of the bugs solved in this PR

Cargo.toml
```diff
@@ -237,4 +237,3 @@ polimec-runtime = { path = "runtimes/polimec" }
 asset-hub-polkadot-runtime = { git = 'https://github.com/polkadot-fellows/runtimes', default-features = false, rev = '59172d5fb8feea2646490618bead3b2b4fb06220' }
 polkadot-runtime-constants = { git = 'https://github.com/polkadot-fellows/runtimes', default-features = false, rev = '59172d5fb8feea2646490618bead3b2b4fb06220' }
 system-parachains-constants = { git = 'https://github.com/polkadot-fellows/runtimes', default-features = false, rev = '59172d5fb8feea2646490618bead3b2b4fb06220' }
-polkadot-runtime = { git = 'https://github.com/polkadot-fellows/runtimes', default-features = false, rev = '59172d5fb8feea2646490618bead3b2b4fb06220' }
```
vstam1, line 240: Wait, is the dep now removed? Should still keep it for integration tests, right?

runtimes/polimec/src/lib.rs
vstam1: Hmmm interesting that it works with Here, as we should not be able to reserve_based transfer PLMC.

vstam1 (approved):
After comments, I approve

Cargo.toml
JuaniRios: Thought your comment was to delete this line?

vstam1: I think in the previous version is added a new line after this line. So was a nit to remove the new line. But we should probably still include this line, else integration tests won't compile

PR #250, opened 2024-04-19
chore: use production in Docker Image

lrazovic:
## What?

+ Use the `production` binary in the Docker Image.
+ The chainspec is not needed anymore since it's part of the node now.


JuaniRios (approved):
lgtm

PR #251, opened 2024-04-19
🚑️ Use only the WasmExecutor

lrazovic:
## What?

+ The node should only allow the WASM execution: https://github.com/encointer/encointer-parachain/pull/221/

## Why?

+ After the latest runtime upgrade we encountered several synchronisation problems between nodes because some nodes used a native runtime different from the canonical one (the WASM one). 

> the native runtime is compiled natively into the node and brings some better performance than the wasm runtime. However, when you upgrade the runtime on chain you always send the wasm runtime. The wasm runtime is the canonical one. 
>
> src: https://substrate.stackexchange.com/a/153/3722


nodes/parachain/Cargo.toml
```diff
@@ -49,14 +48,14 @@ sc-transaction-pool.workspace = true
 sc-transaction-pool-api.workspace = true
 sc-network-sync.workspace = true
 sp-api.workspace = true
+sp-io = { workspace = true }
```
JuaniRios, line 51:
```suggestion
sp-io.workspace = true
```

JuaniRios (approved):
Small NIT.

Can't judge the logic of this change since I'm not familiar with the node logic, but I guess we trust Basti's fix in Encointer per the link given. This seems 1 to 1 their change

vstam1 (approved):
Like Juan said, do not 100% understand the logic, but looks like 1 to 1 with encointer. What is the best way to test this change? Running new node? Upgrading our nodes?

vstam1:
Alright lets merge and keep track of the upgrade process to see if we encounter problems!

PR #258, opened 2024-04-24
Leo/restrict token audiance

lrazovic:
## What
- Add project policy checks for evaluation/participation extrinsics

## Why
- Missing feature for MVP

## How
- Compare the policy in the project metadata with the one coming from the JWT

## Testing
- One added test in each evaluation/participation module

### DIFF
--- a/pallets/dispenser/src/benchmarking.rs
+++ b/pallets/dispenser/src/benchmarking.rs
@@ -33,6 +33,7 @@ fn assert_last_event<T: Config>(generic_event: <T as Config>::RuntimeEvent) {
 #[benchmarks]
 mod benchmarks {
 	use super::*;
+	use polimec_common_test_utils::get_mock_jwt_with_cid;
 
 	#[benchmark]
 	fn dispense() {
--- a/pallets/dispenser/src/benchmarking.rs
+++ b/pallets/dispenser/src/benchmarking.rs
@@ -33,8 +33,6 @@ fn assert_last_event<T: Config>(generic_event: <T as Config>::RuntimeEvent) {
 #[benchmarks]
 mod benchmarks {
 	use super::*;
-	use polimec_common_test_utils::get_mock_jwt_with_cid;
-
 	#[benchmark]
 	fn dispense() {
 		let caller: T::AccountId = whitelisted_caller();
--- a/pallets/funding/src/lib.rs
+++ b/pallets/funding/src/lib.rs
@@ -805,14 +805,10 @@ pub mod pallet {
 			project_id: ProjectId,
 			#[pallet::compact] usd_amount: BalanceOf<T>,
 		) -> DispatchResultWithPostInfo {
-			let (account, did, investor_type, cid) =
+			let (account, did, investor_type, whitelisted_policy) =
 				T::InvestorOrigin::ensure_origin(origin, &jwt, T::VerifierPublicKey::get())?;
 
-			// JUST FOR DEMO PURPOSES
-			let project = ProjectsMetadata::<T>::get(project_id).unwrap();
-			// Create a new Error to comunicate that the used JWT is not validated against this project's policy.
-			ensure!(project.policy_ipfs_cid == Some(cid), Error::<T>::BadMetadata(MetadataError::CidNotProvided));
-			Self::do_evaluate(&account, project_id, usd_amount, did, investor_type)
+			Self::do_evaluate(&account, project_id, usd_amount, did, investor_type, whitelisted_policy)
 		}
 
 		/// Bid for a project in the Auction round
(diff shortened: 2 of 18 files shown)


pallets/funding/src/lib.rs
```diff
@@ -805,8 +805,13 @@ pub mod pallet {
...
 			#[pallet::compact] usd_amount: BalanceOf<T>,
 		) -> DispatchResultWithPostInfo {
-			let (account, did, investor_type) =
+			let (account, did, investor_type, cid) =
 				T::InvestorOrigin::ensure_origin(origin, &jwt, T::VerifierPublicKey::get())?;
+
+			// JUST FOR DEMO PURPOSES
+			let project = ProjectsMetadata::<T>::get(project_id).unwrap();
+			// Create a new Error to comunicate that the used JWT is not validated against this project's policy.
+			ensure!(project.policy_ipfs_cid == Some(cid), Error::<T>::BadMetadata(MetadataError::CidNotProvided));
```
lrazovic, line 814: We get the CID from the JWT and we check it against the policy

pallets/funding/src/lib.rs
```diff
@@ -734,7 +734,7 @@ pub mod pallet {
 			jwt: UntrustedToken,
 			project: ProjectMetadataOf<T>,
 		) -> DispatchResult {
-			let (account, did, investor_type) =
+			let (account, did, investor_type, _cid) =
```
lrazovic, line 737: Not a big fan of this pattern, but let's improve this later

pallets/funding/src/lib.rs
```diff
@@ -805,8 +805,13 @@ pub mod pallet {
 			project_id: ProjectId,
 			#[pallet::compact] usd_amount: BalanceOf<T>,
 		) -> DispatchResultWithPostInfo {
-			let (account, did, investor_type) =
+			let (account, did, investor_type, cid) =
 				T::InvestorOrigin::ensure_origin(origin, &jwt, T::VerifierPublicKey::get())?;
+
+			// JUST FOR DEMO PURPOSES
+			let project = ProjectsMetadata::<T>::get(project_id).unwrap();
```
JuaniRios, line 812:
```suggestion
			let project_metadata = ProjectsMetadata::<T>::get(project_id).unwrap();
```

polimec-common/common/src/credentials/mod.rs
```diff
@@ -57,20 +57,23 @@ pub struct SampleClaims<AccountId> {
...
+	#[serde(rename = "aud", deserialize_with = "from_bounded_cid")]
+	pub ipfs_cid: Cid,
 	pub investor_type: InvestorType,
-	#[serde(deserialize_with = "from_bounded_vec")]
+	#[serde(deserialize_with = "from_bounded_did")]
 	pub did: Did,
 }
 
 pub type Did = BoundedVec<u8, ConstU32<57>>;
+pub type Cid = BoundedVec<u8, ConstU32<96>>;
```
JuaniRios, line 68: A `BoundedString` is the same inner type as a `BoundedVec` no?

polimec-common/test-utils/src/lib.rs
```diff
@@ -60,26 +67,41 @@ pub fn get_mock_jwt<AccountId: frame_support::Serialize>(
...
+// The `Serialize` trait is needed to serialize the `account_id` into a  `SampleClaims` struct.
+pub fn get_mock_jwt_with_cid<AccountId: frame_support::Serialize>(
+	account_id: AccountId,
+	investor_type: InvestorType,
+	did: BoundedVec<u8, ConstU32<57>>,
+	ipfs_cid: BoundedVec<u8, ConstU32<96>>,
+) -> UntrustedToken {
+	create_jwt(account_id, investor_type, did, Some(ipfs_cid))
+}
+
```
JuaniRios, line 104: Why not have just one function where ipfs_cid is an `Option<Cid>`?

lrazovic: That was my initial approach, but it caused all the tests to fail. To avoid modifying all the tests at this early stage of implementation, I opted for a different function

pallets/dispenser/src/benchmarking.rs
```diff
@@ -33,6 +33,7 @@ fn assert_last_event<T: Config>(generic_event: <T as Config>::RuntimeEvent) {
 #[benchmarks]
 mod benchmarks {
 	use super::*;
+	use polimec_common_test_utils::get_mock_jwt_with_cid;
```
vstam1, line 36: Think this breaks the dispenser benchmarks as code is still using get_mock_jwt or it is an unnecessary import?

vstam1 (approved):
Lgtm, only one small question about dispenser benchmarks.

pallets/dispenser/src/benchmarking.rs
lrazovic: It was an unnecessary import https://github.com/Polimec/polimec-node/pull/258/commits/fcac9f47fab3a9d0941382729793227a17e6da16

PR #260, opened 2024-04-25
💥 New decimal-aware price system - Part 1: Business Logic

JuaniRios:
## What?
- Introduce a new price system so we can convert `asset->usd->asset` correctly when they have different decimals.

## Why?
- Even if issuers use 10 decimals for their CT, we already know USDT and USDC use 6 decimals, so any conversions would be wrong.

## How?
Change the price of assets instead of being for example PriceOf::from_float(1.5) for a price of 1.5$, we now use a relationship between our USD decimals and the asset decimals like so `PriceOf::from_rational(1.5 * 10^6, 1 * 10^10)` for the case where our base USD decimals is 6, and the asset being priced has 10 decimals

## Testing?
`test_usd_price_decimal_aware`
All other tests are broken and will be fixed in a following PR on the stack


JuaniRios:
* **#260** <a href="https://app.graphite.dev/github/pr/Polimec/polimec-node/260?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

pallets/funding/src/traits.rs
```diff
@@ -32,6 +32,22 @@ pub trait ProvideAssetPrice {
 	type AssetId;
 	type Price: FixedPointNumber;
 	fn get_price(asset_id: Self::AssetId) -> Option<Self::Price>;
+
+	fn calculate_decimals_aware_price(
+		original_price: Self::Price,
+		usd_decimals: u8,
+		asset_decimals: u8,
+	) -> Option<Self::Price> {
+		let usd_unit = 10u128.checked_pow(usd_decimals.into())?;
```
vstam1, line 41: Should probably be one? Now usd_unit is 10_000_000? 
Same for `asset_unit`.

pallets/funding/src/traits.rs
```diff
@@ -32,6 +32,22 @@ pub trait ProvideAssetPrice {
 	type AssetId;
 	type Price: FixedPointNumber;
 	fn get_price(asset_id: Self::AssetId) -> Option<Self::Price>;
+
+	fn calculate_decimals_aware_price(
```
vstam1, line 36: Could you add some comments with an calculation example explaining what this function does? Also for get_decimals_aware_price?

pallets/funding/src/functions.rs
```diff
@@ -1405,9 +1425,15 @@ impl<T: Config> Pallet<T> {
...
+		let plmc_usd_price = T::PriceProvider::get_decimals_aware_price(PLMC_FOREIGN_ID, USD_DECIMALS, PLMC_DECIMALS)
+			.ok_or(Error::<T>::PriceNotFound)?;
+
+		let funding_asset_id = funding_asset.to_assethub_id();
+		let funding_asset_decimals = T::FundingCurrency::decimals(funding_asset_id);
 		let funding_asset_usd_price =
-			T::PriceProvider::get_price(funding_asset.to_assethub_id()).ok_or(Error::<T>::PriceNotFound)?;
+			T::PriceProvider::get_decimals_aware_price(funding_asset_id, USD_DECIMALS, funding_asset_decimals)
+				.ok_or(Error::<T>::PriceNotFound)?;
+
```
vstam1, line 1436: Could choose to add these three steps in a specific funding pallet function as we will be using it often:
getting asset id,
getting decimals
getting decimal aware price.

pallets/funding/src/tests/misc.rs
```diff
@@ -4,6 +4,21 @@ use super::*;
 mod helper_functions {
 	use super::*;
 
+	#[test]
+	fn test_usd_price_decimal_aware() {
+		let submitted_price = FixedU128::from_float(1.85);
+		let asset_decimals = 4;
+		let expected_price = FixedU128::from_float(185.0);
+
+		assert_eq!(PolimecFunding::usd_price_decimal_aware(submitted_price, asset_decimals).unwrap(), expected_price);
```
vstam1, line 13: Function does not exist?

vstam1 (commented):
Think its a good first setup. Some small things, hope that testing will be easy!

pallets/funding/src/traits.rs
JuaniRios: 10^6 = 1_000_000
1 ^ 6 = 1

So this is correct, or am I missing something?

pallets/funding/src/tests/misc.rs
JuaniRios: Fixed

pallets/funding/src/functions.rs
JuaniRios: We only use it bid(), and contribute(). I prefer to keep it verbose for now and abstract it if we keep running into it later. I'll keep this in mind

pallets/funding/src/traits.rs
vstam1: Oh wow, my math head let me down here. You are completely right!

pallets/funding/src/traits.rs
```diff
@@ -33,6 +33,11 @@ pub trait ProvideAssetPrice {
 	type Price: FixedPointNumber;
 	fn get_price(asset_id: Self::AssetId) -> Option<Self::Price>;
 
+	/// Prices define the relationship between USD/Asset. When to and from that asset, we need to be aware that they might
+	/// have different decimals. This function calculates the relationship having in mind the decimals. For example:
+	/// if the price is 2.5, our underlying USD unit has 6 decimals, and the asset has 8 decimals, the price will be
+	/// calculated like so: `(2.5USD * 10^6) / (1 * 10^8) = 0.025`. And so if we want to convert 20 of the asset to USD, we would do
+	/// we would do `0.025(USD/Asset)FixedPointNumber * 20_000_000_00(Asset)u128 = 50_000_000` which is 50 USD with 6 decimals
```
vstam1, line 40:
```suggestion 
	/// calculated like so: `(2.5USD * 10^6) / (1 * 10^8) = 0.025`. And so if we want to convert 20 of the asset to USD,
	/// we would do `0.025(USD/Asset)FixedPointNumber * 20_000_000_00(Asset)u128 = 50_000_000` which is 50 USD with 6 decimals

File: pallets/funding/src/types.rs [hunk 6]
```diff
@@ -285,7 +285,7 @@ pub mod storage_types {
 		/// When the different project phases start and end
 		pub phase_transition_points: PhaseTransitionPoints<BlockNumber>,
 		/// Fundraising target amount in USD equivalent
```
vstam1, line 287: Please at to this comment, that USD equivalent means 6 decimal USD.

vstam1 (approved):

lrazovic (approved):

pallets/funding/src/types.rs
```diff
@@ -284,12 +284,12 @@ pub mod storage_types {
 		pub status: ProjectStatus,
 		/// When the different project phases start and end
 		pub phase_transition_points: PhaseTransitionPoints<BlockNumber>,
-		/// Fundraising target amount in USD equivalent
-		pub fundraising_target: Balance,
+		/// Fundraising target amount in USD (6 decimals)
+		pub fundraising_target_usd: Balance,
```
lrazovic, line 288: @felixpolimec: Rename from `fundraising_target` to `fundraising_target_usd` in `ProjectDetails`.

pallets/funding/src/types.rs
```diff
@@ -284,12 +284,12 @@ pub mod storage_types {
...
 		/// The amount of Contribution Tokens that have not yet been sold
 		pub remaining_contribution_tokens: Balance,
-		/// Funding reached amount in USD equivalent
-		pub funding_amount_reached: Balance,
+		/// Funding reached amount in USD (6 decimals)
+		pub funding_amount_reached_usd: Balance,
```
lrazovic, line 292: @felixpolimec: Rename from `funding_amount_reached` to `funding_amount_reached_usd` in `ProjectDetails`.

PR #273, opened 2024-04-30
🔧 Slow down the rounds in fast-mode

lrazovic:
## What and Why?

Increase the rounds duration to 1 hour, so @felixpolimec has enough time to test the various rounds.


JuaniRios (approved):
lgtm

PR #281, opened 2024-05-02
Make sure USD is representable with CT decimal/price combo

JuaniRios:
## What?
- Find ways to avoid unrepresentability of issuer-set tokenomics on our own types

## Why?
- We let issuers choose whatever price, allocation, and decimals they want, but we still need to convert those into our own FixedU128 price, and 6 decimal USD amounts. 
- This could potentially lead to some issuer-set tokenomics being unrepresentable with our own types, For example due to their CT being too high in value, and we couldn't represent 1 US cent with their lowest possible CT amount, or their price decimals having more decimals than our Price type, and it saturating to 0.

## How?
- Limit the decimals to 6-18 range, and price to 0.00001-1000
- After experimenting some on the misc tests, I came to the conclusion that the best course now is to severely restrict the decimals and price that the issuer can put. 
- We will need a lot more testing and consultation with other experts if we want to increase those limits. 
- A potential solution would be to change how we handle prices and decimals, like using for example something like  https://github.com/sam0x17/currencies/tree/main

## Testing?
- New tests were written on the application file to check for the different decimals and price limits

## Anything else?
- Also fixed broken tests from main
- We need another PR were we replace all unsafe `pow` with safe `checked_pow`

### DIFF
--- a/pallets/funding/src/tests/misc.rs
+++ b/pallets/funding/src/tests/misc.rs
@@ -674,4 +674,35 @@ mod bug_hunting {
 		let new_project_details = inst.get_project_details(project_id);
 		assert_eq!(old_project_details, new_project_details);
 	}
+
+	// This shows us that "plausible" combination of ct decimals and ct price can lead to unrepresentable usd cents
+	#[test]
+	fn price_too_low_for_usd_conversion_1() {
+		let ct_decimals = 4u8;
+		let original_price = FixedU128::from_float(1000f64);
+		let decimal_aware_price = <TestRuntime as Config>::PriceProvider::calculate_decimals_aware_price(
+			original_price,
+			USD_DECIMALS,
+			ct_decimals,
+		)
+		.unwrap();
+
+		let usd_cent = USD_UNIT / 100;
+		let cent_in_ct = decimal_aware_price.reciprocal().unwrap().checked_mul_int(usd_cent).unwrap();
+		assert_eq!(cent_in_ct, 0u128);
+	}
+
+	// This shows us that even using the lowest possible price with FixedU128, we can still represent 1bn USD
+	#[test]
+	fn billion_saturating() {
+		let decimal_aware_price = FixedU128::from_inner(1u128);
+		dbg!(decimal_aware_price);
+
+		let usd_billion = USD_UNIT * 1_000_000_000;
+		let billion_in_ct = decimal_aware_price.reciprocal().unwrap().checked_mul_int(usd_billion).unwrap();
+		dbg!(billion_in_ct);
+		dbg!(billion_in_ct.checked_ilog10());
+		dbg!(u128::MAX);
+		dbg!(u128::MAX.checked_ilog10());
+	}
 }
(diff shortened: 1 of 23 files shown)


JuaniRios:
* **#281** <a href="https://app.graphite.dev/github/pr/Polimec/polimec-node/281?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

pallets/funding/src/lib.rs
```diff
@@ -235,7 +235,7 @@ pub mod pallet {
 		type AuctionInitializePeriodDuration: Get<BlockNumberFor<Self>>;
 
 		/// The inner balance type we will use for all of our outer currency types. (e.g native, funding, CTs)
-		type Balance: Balance + From<u64> + FixedPointOperand + MaybeSerializeDeserialize + Into<u128>;
+		type Balance: Balance + From<u64> + FixedPointOperand + MaybeSerializeDeserialize + Into<u128> + Bounded;
```
vstam1, line 238: Where do we need the Bounded traitbound for?

pallets/funding/src/tests/misc.rs
```diff
@@ -674,4 +674,36 @@ mod bug_hunting {
...
+		let decimal_aware_price = FixedU128::from_inner(1u128);
+		dbg!(decimal_aware_price);
+
+		let usd_billion = USD_UNIT * 1_000_000_000;
+		let billion_in_ct = decimal_aware_price.reciprocal().unwrap().checked_mul_int(usd_billion).unwrap();
+		dbg!(billion_in_ct);
+		dbg!(billion_in_ct.checked_ilog10());
+		dbg!(u128::MAX);
+		dbg!(u128::MAX.checked_ilog10());
+	}
```
vstam1, line 708: If we want to keep these tests, can we move from dbg to asserts. Else please remove as they are not actually "testing" anything.

vstam1 (commented):
Think the logic is solid, but do not completely comprehend it. Couple of comments on rust/tests, but think the minimum price bounds look good. You are write documentation by the create project extrinsic/metadata struct about what each value represents so that its properly documented in the code. So what is the input of our create_project extrinsic and how does it compute to the on-chain storage values. Would be highly appriciated

pallets/funding/src/lib.rs
JuaniRios: Good catch, it was needed in a previous version where we checked for close to overflow of Balance.

pallets/funding/src/types.rs
```diff
@@ -200,7 +195,7 @@ pub mod storage_types {
 
 	impl<
 			BoundedString,
-			Balance: From<u64> + PartialOrd + Copy + FixedPointOperand,
+			Balance: From<u64> + PartialOrd + Copy + FixedPointOperand + One + CheckedMul,
```
vstam1, line 198: Do we need One traitbound?

vstam1 (approved):
Single comment

PR #289, opened 2024-05-07
:sparkles: add filter on Funding calls

vstam1:
## What?
Add filter on Funding calls.
Only allow:
create, edit and remove project calls.


JuaniRios (approved):
lgtm. To be releasable we need to increase the polimec version though

vstam1:
Are we not already on 0.7.0? Live version is 0.6.0 so think it should be fine.

lrazovic (approved):

PR #290, opened 2024-05-08
feat: set the production verifier public key

lrazovic:
## What?

Set the `VerifierPublicKey` to a public key matching the private key used to generate the JWT in production.

### DIFF
--- a/runtimes/polimec/src/lib.rs
+++ b/runtimes/polimec/src/lib.rs
@@ -982,7 +982,8 @@ parameter_types! {
 
 }
 
-#[cfg(not(feature = "on-chain-release-build"))]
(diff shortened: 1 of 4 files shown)


runtimes/polimec/src/lib.rs
vstam1, line 1: Can you include the change for pallet-funding PalletId from py/fund to plmc/fun (or something similar)

lrazovic: https://github.com/Polimec/polimec-node/pull/290/commits/e49d87733e616c405e48fd49a42e4608f8719406

vstam1 (approved):

PR #296, opened 2024-05-14
🎨 Restructure funding functions file

JuaniRios:
## What?
- Restructure all functions in `functions.rs` into separate submodules

## Why?
- Better readability

## How?
Use a similar separation as we did in tests

## Testing?
Normal tests run

PR #299, opened 2024-05-16
✨ Optimize wap calculation

JuaniRios:
## What?
- Spread out the WAP on_initialize calculation into 2 blocks

## Why?
- Because the ref time was too close to the block limit

## How?
- Create a do_end_auction_closing function that does the deciding of which bids are successful, and refunds/deletes failed ones
- do_start_community now does the WAP calculation only

## Testing?
- Run normal tests, all logic should be maintained

### DIFF
--- a/pallets/funding/src/functions/3_auction.rs
+++ b/pallets/funding/src/functions/3_auction.rs
@@ -4,7 +4,7 @@ impl<T: Config> Pallet<T> {
 	/// Called by user extrinsic
 	/// Starts the auction round for a project. From the next block forward, any professional or
 	/// institutional user can set bids for a token_amount/token_price pair.
-	/// Any bids from this point until the auction_closing starts, will be considered as valid.
+	/// Any bids from this point until the auction_closing starts will be considered as valid.
 	///
 	/// # Arguments
 	/// * `project_id` - The project identifier
@@ -24,7 +24,7 @@ impl<T: Config> Pallet<T> {
 	/// Later on, `on_initialize` transitions the project into the closing auction round, by calling
 	/// [`do_auction_closing`](Self::do_auction_closing).
 	#[transactional]
-	pub fn do_auction_opening(caller: AccountIdOf<T>, project_id: ProjectId) -> DispatchResultWithPostInfo {
+	pub fn do_start_auction_opening(caller: AccountIdOf<T>, project_id: ProjectId) -> DispatchResultWithPostInfo {
 		// * Get variables *
 		let mut project_details = ProjectsDetails::<T>::get(project_id).ok_or(Error::<T>::ProjectDetailsNotFound)?;
 		let now = <frame_system::Pallet<T>>::block_number();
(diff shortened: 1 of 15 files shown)


JuaniRios:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/Polimec/polimec-node/299?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>
> <a href="https://graphite.dev/docs/merge-pull-requests">Learn more</a>

* **#299** <a href="https://app.graphite.dev/github/pr/Polimec/polimec-node/299?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#296** <a href="https://app.graphite.dev/github/pr/Polimec/polimec-node/296?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

pallets/funding/src/functions/3_auction.rs
```diff
@@ -149,6 +149,68 @@ impl<T: Config> Pallet<T> {
 		})
 	}
 
+	#[transactional]
+	pub fn do_end_auction_closing(project_id: ProjectId) -> DispatchResultWithPostInfo {
+		// * Get variables *
+		let mut project_details = ProjectsDetails::<T>::get(project_id).ok_or(Error::<T>::ProjectDetailsNotFound)?;
```
lrazovic, line 155:
```suggestion
		let project_details = ProjectsDetails::<T>::get(project_id).ok_or(Error::<T>::ProjectDetailsNotFound)?;
```

lrazovic (changes requested):
Few nits here and there, but the logic looks good! Let's merge after the approval from @vstam1 since he was more involved in the design of this.

pallets/funding/src/types.rs
```diff
@@ -775,6 +780,18 @@ pub mod inner_types {
...
+	#[derive(Clone, Encode, Decode, Eq, PartialEq, RuntimeDebug, TypeInfo, MaxEncodedLen)]
+	pub struct AuctionRoundInfo<Balance> {
+		// Will be `Some` after the auction closing phase ends and winning bids are calculated
+		pub is_oversubscribed: IsOversubscribed<Balance>,
+	}
+	#[derive(Clone, Encode, Decode, Eq, PartialEq, RuntimeDebug, TypeInfo, MaxEncodedLen)]
+	pub enum IsOversubscribed<Balance> {
+		Yes { total_usd_bid: Balance },
+		No,
+	}
```
vstam1, line 793: I think I would prefer an Option here instead of the IsOversubscribed enum. I think you had an option as well looking at the comments.

vstam1: You could rename the AuctionRoundInfo (as it only holds the is_oversubscribed inner value). So we sort of unpack the inner value as the outer value, e.g. AuctionRoundInfo -> IsOversubscribed (or some other name), which is an option. If we add a clear doc comment on the value, I think it would be clear enough what the Some(value) means.

vstam1: So instead of
`pub auction_round_info: AuctionRoundInfo,`
we do:
```
/// Whether or not the auction is oversubscribed. An auction is oversubscribed if the first bucket has been filled. 
/// In this case we have to calculate the weighted average price.
/// If the auction is oversubscribed we store the usd total of all accepted bids.
/// We use this value for the wap calculation. 
pub auction_oversubscribed: Option<Balance>,
```

vstam1 (approved):

PR #317, opened 2024-06-05
✨ Paseo config

JuaniRios:
## What?
- Add paseo chain spec
- Refactor all chain specs

## Why?
- To start our paseo testnet
- To avoid duplicating code and improve readability

## How?
- Split chain specs into files depending on their relay chain
- Remove unused live chain specs, keep the json files at ./chain-specs

## Testing?
- Run with zombienet each chain spec and check if we produce blocks


nodes/parachain/src/chain_spec/common.rs
```diff
@@ -0,0 +1,193 @@
+use crate::chain_spec::{get_account_id_from_seed, get_properties, Extensions, GenericChainSpec, DEFAULT_PARA_ID};
+use cumulus_primitives_core::ParaId;
+use frame_support::traits::fungible::Inspect;
+use itertools::Itertools;
```
vstam1, line 4: Why do we need the Itertools again?

scripts/zombienet/polimec-polkadot-assethub-penpal.toml
```diff
@@ -0,0 +1,111 @@
...
+
+
+    [[parachains.collators]]
+    name = "collator2"
+    command = "./bin/polimec-node"
+    args = ["--offchain-worker when-authority", "-lerror", "-lxcm::process_instruction=trace", "-lxcm::execute_xcm_in_credit=trace", "-lxcm::barriers=trace", "-lpallet_funding::hrmp=trace", "-lpallet_funding::test=trace", "--execution=native"]
+    # ss58 key: 5E5E37FNZD9KVHyGgSHt8pc2kq8e3VUS5rf8GmrxCa7ySs8s
+    keystore_key_types = ["aura", "plmc_sr"]
+
+    [[parachains.collators]]g
```
vstam1, line 59: Sure this works? There is some random 'g' here at the end

scripts/zombienet/polimec-polkadot-assethub-penpal.toml
```diff
@@ -0,0 +1,111 @@
...
+    ws_port = 7777
+
+
+[[parachains]]
+id = 3344
+chain = "polimec-polkadot-local"
+    [[parachains.collators]]
+    name = "collator1"
+    command = "./target/release/polimec-node"
+    args = ["--offchain-worker when-authority", "-lerror", "-lxcm::process_instruction=trace", "-lxcm::execute_xcm_in_credit=trace", "-lxcm::barriers=trace", "-lpallet_funding::hrmp=trace", "-lpallet_funding::test=trace", "--execution=native"]
```
vstam1, line 46: Maybe only use the args for 1 collator? We will probably only check the logs with a single collator. Let me know if you think otherwise

vstam1 (commented):
Think in general it looks pretty good! Couple of comments. Do think its good if Leo has some time to take a second look, as I'm no node side expert, so might not have seen some things.

nodes/parachain/src/chain_spec/common.rs
JuaniRios: `collect_vec`

vstam1 (approved):
Lgtm, but Leo should definitely do a review as well!

nodes/parachain/Cargo.toml
```diff
@@ -74,6 +74,7 @@ cumulus-client-consensus-proposer.workspace = true
 cumulus-client-service.workspace = true
 cumulus-primitives-core.workspace = true
 cumulus-relay-chain-interface.workspace = true
+itertools = "0.10.5"
```
lrazovic, line 77: Let's use the workspace https://github.com/Polimec/polimec-node/blob/205b7c41dd7c7f5ba9927694397e49d5bcf10f91/Cargo.toml#L88

nodes/parachain/src/command.rs
```diff
@@ -86,30 +86,30 @@ fn runtime(id: &str) -> Runtime {
...
-		// Base runtime
-		"polimec" => Box::new(chain_spec::polimec::ChainSpec::from_json_bytes(polimec_chain_spec)?),
-		"rolimec" => Box::new(chain_spec::polimec::ChainSpec::from_json_bytes(rolimec_chain_spec)?),
-		"politest" => Box::new(chain_spec::politest::get_live_chain_spec()),
-		"polimec-local" => Box::new(chain_spec::polimec::get_local_chain_spec()),
-		"rolimec-local" => Box::new(chain_spec::polimec::get_rococo_chain_spec()),
-		"politest-local" => Box::new(chain_spec::politest::get_local_chain_spec()),
-		// -- Fallback (generic chainspec)
+		// Original live Polimec runtimes
+		"polimec" => Box::new(chain_spec::common::ChainSpec::from_json_bytes(polimec_chain_spec)?),
```
lrazovic, line 94: Can we please add the raw chainspec in the node so we can start a collator by using something like `--chain polimec-paseo`?

lrazovic (changes requested):
We are ready to take the leap to Paseo!

Adding the raw chainspec to the node would be a huge help when it comes to setting up collators

lrazovic (approved):

PR #327, opened 2024-06-19
👨🏻‍💻 Allow polimec receiver pallet to have any index

JuaniRios:
## What?
- Allow parachains to have their polimec receiver pallet assigned to any pallet index

## Why?
- We should not opinionate that

## How?
- We no longer have the config item `PolimecReceiverInfo`. 
- We previously needed it because the `PalletInfo` items returned on the pallet query were set to private. Now we can just check the pallet name and save the index for later

## Testing?
- Modify penpal so the receiver has a different index and run the integration ct_migration tests

### DIFF
--- a/integration-tests/penpal/src/lib.rs
+++ b/integration-tests/penpal/src/lib.rs
@@ -672,8 +672,8 @@ construct_runtime!(
 
 		// The main stage.
 		Assets: pallet_assets::{Pallet, Call, Storage, Event<T>} = 50,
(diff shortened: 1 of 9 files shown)


JuaniRios:
* **#327** <a href="https://app.graphite.dev/github/pr/Polimec/polimec-node/327?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

vstam1 (approved):

PR #328, opened 2024-06-19
fix wrong fallback weight in on_initialize

vstam1:
## What?
This pr adds the `root_do_end_auction_closing` extrinsic and uses this extrinsics weight as fallback in on_initialize for the do_end_auction_closing, that is currently set incorrectly

### DIFF
--- a/pallets/funding/src/lib.rs
+++ b/pallets/funding/src/lib.rs
@@ -1164,14 +1164,14 @@ pub mod pallet {
 			Self::do_evaluation_end(project_id)
 		}
 
-		#[pallet::call_index(59)]
+		#[pallet::call_index(29)]
 		#[pallet::weight(WeightInfoOf::<T>::start_auction_manually(<T as Config>::MaxProjectsToUpdateInsertionAttempts::get() - 1))]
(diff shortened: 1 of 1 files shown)


pallets/funding/src/lib.rs
```diff
@@ -1232,7 +1253,7 @@ pub mod pallet {
 			Self::do_end_funding(project_id)
 		}
 
-		#[pallet::call_index(33)]
+		#[pallet::call_index(36)]
```
JuaniRios, line 1256: Shouldn't this be 35, and then 36?

JuaniRios (changes requested):

pallets/funding/src/lib.rs
vstam1: We added `remove_project` as 35. So I see three solutions:
1. Keep it like this, so 36 instead of 35
2. Change index of remove_project to some other value
3. actually reindex everything but not sure if that will break things.

JuaniRios: I would reindex everything. Better do it know when there's not any big repercusions @lrazovic ?

JuaniRios (approved):

lrazovic (approved):

PR #338, opened 2024-06-25
Set migration user to Multilocation

JuaniRios:


JuaniRios:
* **#338** <a href="https://app.graphite.dev/github/pr/Polimec/polimec-node/338?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

pallets/polimec-receiver/src/lib.rs
```diff
@@ -31,6 +31,7 @@ pub mod pallet {
 	use polkadot_parachain_primitives::primitives::{Id as ParaId, Sibling};
 	use sp_runtime::traits::{AccountIdConversion, Convert};
 	use sp_std::prelude::*;
+	use xcm::v3::{Junction::AccountId32, Junctions::X1, MultiLocation};
```
vstam1, line 34: Maybe good to set this to xcm::latest as well as you are using it on line 64

polimec-common/common/src/lib.rs
```diff
@@ -101,10 +101,11 @@ pub trait ReleaseSchedule<AccountId, Reason> {
 
 pub mod migration_types {
 	use super::*;
+	use xcm::v3::MultiLocation;
```
vstam1, line 104: Also use latest instead of v3?

pallets/funding/src/functions/6_settlement.rs
```diff
@@ -448,8 +448,11 @@ impl<T: Config> Pallet<T> {
 		vesting_time: BlockNumberFor<T>,
 	) -> DispatchResult {
 		UserMigrations::<T>::try_mutate(project_id, origin, |maybe_migrations| -> DispatchResult {
-			let migration_origin =
-				MigrationOrigin { user: T::AccountId32Conversion::convert(origin.clone()), id, participation_type };
+			let multilocation_user = MultiLocation::new(
+				0,
+				X1(AccountId32 { network: None, id: T::AccountId32Conversion::convert(origin.clone()) }),
+			);
```
vstam1, line 454: Could choose to create the ML via tuple().into() implementation (see [example](https://wiki.polkadot.network/docs/learn/xcm/fundamentals/multilocation-example))

vstam1 (commented):
I think it looks good in general. Only thing I'm thinking about is how to work with the different xcm versions. XCMv4 is of course now created, and I think most of the XCM types have changed a bit (not sure about their encoding though). But if we are storing them as XCMv3 now, we might need to do a migration when we move to XCMv4 (or 5)

JuaniRios:
@vstam1 good point, but I think the migration should be fairly trivial no?

vstam1 (approved):

PR #350, opened 2024-07-04
♻️ Summer cleanup

lrazovic:
## What?

Chores' PR to fix various problems identified by `clippy` and `zepter`. It also removes the unused `futures` dependency from pallet-funding and the now unused `PreImageLimit` config type.


JuaniRios (approved):
Good cleanup 🧹

vstam1 (approved):

PR #354, opened 2024-07-08
😎 Runtime APIs

JuaniRios:
## What?
- Add pallet-funding runtime APIs as requested in https://linear.app/polimec/issue/PLMC-582/runtime-api-to-have-less-logic-in-the-ui
- Add and modify instantiator functions to make the testing easier

## Why?
Complex querying and calculations are done better in Rust with the full on-chain data and types than in Javascript on the front-end

## Testing?
- New file inside test folder "runtime_api.rs"

## Anything Else?
To get an overview of the main changes, check out runtime_api.rs on /src and on /tests folders, as well as the new tests.rs file on the instantiator, where I test the new functions.

The rest of the changed files should be more about reflecting the changes of the instantiator functions

### DIFF
--- a/pallets/funding/src/runtime_api.rs
+++ b/pallets/funding/src/runtime_api.rs
@@ -1,0 +1,230 @@
+use crate::{traits::ProvideAssetPrice, *};
+use alloc::collections::BTreeMap;
+use frame_support::traits::fungibles::{metadata::Inspect as MetadataInspect, Inspect, InspectEnumerable};
+use itertools::Itertools;
+use parity_scale_codec::{Decode, Encode};
+use polimec_common::USD_DECIMALS;
+use scale_info::TypeInfo;
+use sp_runtime::traits::Zero;
+
+#[derive(Debug, Clone, PartialEq, Eq, Encode, Decode, TypeInfo)]
+pub struct ProjectParticipationIds<T: Config> {
+	account: AccountIdOf<T>,
+	evaluation_ids: Vec<u32>,
+	bid_ids: Vec<u32>,
+	contribution_ids: Vec<u32>,
+}
+
+sp_api::decl_runtime_apis! {
+	#[api_version(1)]
+	pub trait Leaderboards<T: Config> {
+		/// Get the top evaluations made for a project by the amount of PLMC bonded
+		fn top_evaluations(project_id: ProjectId, amount: u32) -> Vec<EvaluationInfoOf<T>>;
+
+		/// Get the top bids for a project by the amount of CTs bought.
+		fn top_bids(project_id: ProjectId, amount: u32) -> Vec<BidInfoOf<T>>;
+
+		/// Get the top contributions for a project by the amount of CTs bought.
+		fn top_contributions(project_id: ProjectId, amount: u32) -> Vec<ContributionInfoOf<T>>;
+
+		/// Get the top projects by the absolute USD value raised
+		fn top_projects_by_usd_raised(amount: u32) -> Vec<(ProjectId, ProjectMetadataOf<T>, ProjectDetailsOf<T>)>;
+
+		/// Get the top project by the highest percentage of the target reached
+		fn top_projects_by_usd_target_percent_reached(amount: u32) -> Vec<(ProjectId, ProjectMetadataOf<T>, ProjectDetailsOf<T>)>;
+	}
+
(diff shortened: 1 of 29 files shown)


JuaniRios:
* **#354** <a href="https://app.graphite.dev/github/pr/Polimec/polimec-node/354?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

pallets/funding/src/runtime_api.rs
```diff
@@ -0,0 +1,225 @@
...
+					Some((asset_id, balance))
+				} else {
+					None
+				}
+			})
+			.sorted_by(|a, b| b.1.cmp(&a.1))
+			.collect_vec()
+	}
+
+	pub fn funding_asset_to_ct_amount(
```
lrazovic, line 122: 1. Even though this won't be part of the consensus, it would be better to use `expect` instead of just `unwrap`. This way, the UI can reuse the error message inside the expect.
2. Let's remove the `dbg!` messages.
3. More conceptually: To avoid future mismatches, can we abstract the logic for computing the amount and reuse the same function in both the `do_contribute`/`do_bid` logic and here?

pallets/funding/src/runtime_api.rs
```diff
@@ -0,0 +1,225 @@
...
+				current_bucket.update(ct_to_buy)
+			}
+		}
+
+		ct_amount
+	}
+
+	pub fn all_project_participations_by_did(project_id: ProjectId, did: Did) -> Vec<ProjectParticipationIds<T>> {
+		let evaluations = Evaluations::<T>::iter_prefix((project_id,))
+			.filter(|((_account_id, _evaluation_id), evaluation)| evaluation.did == did)
```
lrazovic, line 171: Not tested. but we should be able to abstract the filtering logic, with something like:

```rust
fn collect_ids_by_did<T, I>(iter: I, did: Did) -> Vec<(AccountIdOf<T>, u32)>
where
    I: Iterator<Item = ((AccountIdOf<T>, u32), Item)>,
    Item: { fn did(&self) -> &Did }
{
    iter
        .filter(|((_account_id, _id), item)| item.did() == &did)
        .map(|((account_id, id), _)| (account_id, id))
        .collect_vec()
}
```

Usage:

```rust
let evaluations = collect_ids_by_did(Evaluations::<T>::iter_prefix((project_id,)), did);
    let bids = collect_ids_by_did(Bids::<T>::iter_prefix((project_id,)), did);
```

EDIT: It should be possible to use it also in the `projects_by_did` API

pallets/funding/src/runtime_api.rs
```diff
@@ -0,0 +1,225 @@
...
+				evaluation_ids,
+				bid_ids,
+				contribution_ids,
+			})
+			.collect();
+
+		output
+	}
+
+	pub fn usd_target_percent_reached(project_id: ProjectId) -> FixedU128 {
```
lrazovic, line 212: To ensure a 1:1 result, shouldn't we reuse the same logic that's used in the `do_end_funding` function? (By abstracting the logic there in a function and calling the same function here and there)

lrazovic (changes requested):
Great addition! The new Rust implementation significantly improves our codebase by allowing us to remove complex logic from the UI, which will lead to better performance and easier maintenance. I particularly appreciate the elegant structure you've used - it really showcases Rust's strengths compared to our previous JavaScript implementation.

I've left some inline comments regarding the documentation of individual APIs and some aspects of the logic. The main point I would like to emphasize is ensuring that we align the logic here with what's used in the pallet logic. This consistency will help ensure that the output accurately reflects on-chain behavior

pallets/funding/src/runtime_api.rs
JuaniRios: in do_contribute/do_bid we have the CT amount and calculate the USD amount. In this API we do the reverse. So we cannot really abstract it

pallets/funding/src/runtime_api.rs
JuaniRios: I would need to implement a method on every item type to extract the did, plus all that generic function. Overkill for a couple simple iterator methods.

Happy to revisit this after the simplification PRs

lrazovic (approved):

PR #376, opened 2024-08-09
💿 AWS runtime weight generation

JuaniRios:
## What?
- Generate weights from AWS benchmark server

## Why?
- We need a specific hardware spec for our weights since Polkadot's validators are bounded by an execution time on a specific hardware.


vstam1 (commented):
Think it looks good in general. Only thing I don't understand is how the PoV size of all three settlement functions are the same with 33831? Do you know why?

JuaniRios:
> Think it looks good in general. Only thing I don't understand is how the PoV size of all three settlement functions are the same with 33831? Do you know why? 



Seems that it's some sort of built-in minimum. The measured ones are different, but the estimated are the same

vstam1 (approved):

PR #399, opened 2024-09-10
1️⃣  One Token Model - New Pallet Proxy Bonding

JuaniRios:
## What?
- Make it possible for users of the funding pallet to avoid having to bond PLMC when participating, and instead pay a fee to have the treasury bond its own PLMC

## Why?
- Participants no longer need to hold any PLMC to participate, especially if we allow paying extrinsics with USDT/C/DOT
- Better UI/UX

## How?
- Add a new pallet called proxy-bonding.
- This pallet exposes a function called `bond_on_behalf_of` which takes the account, bond amount, fee asset id, derivation path (project id), and the hold reason.
  - The function does a transfer_on_hold of the bond amount from the treasury, into a sub account generated using the derivation path, and pays the fee from the account to the sub_account,.
- A function called `set_release_type` decides whether the the fee will go to the configured fee recipient, or will be refunded. Also decides if the bond can be immediately send back to the treasury, or after a set block
- A function `refund_fee` refunds the fee of a certain user/bond if the release type was configured as refunded.
- An extrinsic `transfer_bonds_back_to_treasury` does what the name entails, and checks the release type for when it can be done successfully
- An extrinsic `transfer_fees_to_recipient` does what the name entails, only if the release type is `Locked(n)` and not `Refunded`

## Testing?
One for each release type.

## Anything Else?
If the design is approved by the reviewers, I can include the integration tests in a separate PR after this.


JuaniRios:
* **#399** <a href="https://app.graphite.dev/github/pr/Polimec/polimec-node/399?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

pallets/proxy-bonding/src/lib.rs
```diff
@@ -0,0 +1,197 @@
...
+
+		/// The root id used to derive sub-accounts. These sub-accounts will be used to bond the tokens
+		type RootId: Get<Self::Id>;
+	}
+
+	#[pallet::pallet]
+	pub struct Pallet<T>(_);
+
+	#[derive(Clone, Encode, Decode, Eq, PartialEq, RuntimeDebug, MaxEncodedLen, TypeInfo)]
+	pub enum ReleaseType<BlockNumber> {
```
lrazovic, line 97: Usually for the reasons we use the `pallet::composite_enum` [attribute]([url](https://paritytech.github.io/polkadot-sdk/master/frame_support/pallet_macros/attr.composite_enum.html)). Do we have to use “manual” derive in this case because of the generic parameter?

pallets/proxy-bonding/src/functions.rs
```diff
@@ -0,0 +1,112 @@
...
+	/// Calculate the USD fee in `fee_asset` for bonding `bond_amount` of the native token.
+	/// e.g. if the fee is 1%, native token PLMC, fee_asset USDT, bond_amount 1000 PLMC, PLMC price 0.5USD, USDT price 1USD,
+	/// Then the calculated fee would be 1% * 1000 * 0.5 = 5USD, which is 5 USDT at a price of 1USD.
+	pub fn calculate_fee(bond_amount: BalanceOf<T>, fee_asset: AssetId) -> Result<BalanceOf<T>, &'static str> {
+		let bonding_token_price = T::PriceProvider::get_decimals_aware_price(
+			T::BondingTokenId::get(),
+			T::UsdDecimals::get(),
+			T::BondingTokenDecimals::get(),
+		)
+		.ok_or("Price not available")?;
```
lrazovic, line 27: Is there a specific reason why we use `&'static str` as an error and not a `DispatchError`?

lrazovic (changes requested):
Few nits here and there, but the logic looks good and aligns with what we discussed. Just a few minor issues to iron out. I'll approve this once we have some integration tests in place to verify it works as expected on our runtime.

pallets/proxy-bonding/src/lib.rs
JuaniRios: we can no longer have inner values inside the variants of a composite enum, so I cannot add the `Locked(BlockNumber)` as part of the hold reason. Also this is not really a hold reason, and we cannot have custom composite enums, only HoldReason, FreezeReason, etc

pallets/proxy-bonding/src/functions.rs
```diff
@@ -0,0 +1,114 @@
...
+	) -> Result<(), DispatchError> {
+		let bonding_account: AccountIdOf<T> = T::RootId::get().into_sub_account_truncating(derivation_path);
+		let fee_in_fee_asset = Self::calculate_fee(bond_amount, fee_asset)?;
+		let release_type = Releases::<T>::get(derivation_path, hold_reason).ok_or(Error::<T>::ReleaseTypeNotSet)?;
+
+		ensure!(release_type == ReleaseType::Refunded, Error::<T>::FeeRefundDisallowed);
+
+		// We know this fee token account is existing thanks to the provider reference of the ED of the native asset, so we can fully move all the funds.
+		// FYI same cannot be said of the `account`. We assume they only hold the fee token so their fee asset balance must not go below the min_balance.
+		T::FeeToken::transfer(fee_asset, &bonding_account, account, fee_in_fee_asset, Preservation::Expendable)?;
```
vstam1, line 110: Not sure if this will now fail if `fee_in_fee_asset` is below minimum amount for `account`. If its dusted or if the transaction fails. We could already prevent any problems if we call this `refund_fee` after `release_funding_asset` as we are sure that the contribution amount is already enough.

vstam1 (commented):
Think logic looks good. Couple of comments, but can be merged once comments are addressed

pallets/proxy-bonding/src/functions.rs
```diff
@@ -0,0 +1,114 @@
...
+		hold_reason: T::RuntimeHoldReason,
+	) -> Result<(), DispatchError> {
+		let treasury = T::Treasury::get();
+		let bonding_account: AccountIdOf<T> = T::RootId::get().into_sub_account_truncating(derivation_path);
+		let existential_deposit = <T::BondingToken as fungible::Inspect<T::AccountId>>::minimum_balance();
+
+		let fee_in_fee_asset = Self::calculate_fee(bond_amount, fee_asset)?;
+
+		// Pay the fee from the user to the bonding account. It awaits either a full transfer to the T::FeeRecipient, or a refund to each user
+		T::FeeToken::transfer(fee_asset, &account, &bonding_account, fee_in_fee_asset, Preservation::Preserve)?;
```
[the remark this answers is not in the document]
JuaniRios: I think the transfer will only fail if there is no PLMC ED, and the amount you send is under the min deposit. But I'll write a test to confirm this

pallets/proxy-bonding/src/functions.rs
JuaniRios: Good point, we know for sure the funding ticket is > ED, but we have no guarantees the OTM fee is.
FYI This will be changed in #401

pallets/proxy-bonding/src/functions.rs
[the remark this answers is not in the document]
JuaniRios: I added a test putting the min deposits that we have on mainnet, and we never go under the min deposit with the min ticket fee.

Test is at integration_tests::tests::otm_edge_cases.rs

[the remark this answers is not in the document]
JuaniRios: UPDATE: it will error and storage roll back

lrazovic (approved):

PR #400, opened 2024-09-13
🩹 Update zombienet config

lrazovic:
## What?

Fix zombienet config and a small warning while compiling with the `on-chain-release-build` feature flag.

### DIFF
--- a/scripts/zombienet/polimec-paseo-local.toml
+++ b/scripts/zombienet/polimec-paseo-local.toml
@@ -4,7 +4,7 @@ provider = "native"
 
 [relaychain]
 default_command = "polkadot"
-chain_spec_path = "./scripts/zombienet/relay-chain-specs/paseo-local.plain.json"
+chain = "rococo-local"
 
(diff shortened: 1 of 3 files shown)


scripts/zombienet/polimec-paseo-local.toml
```diff
@@ -4,7 +4,7 @@ provider = "native"
 
 [relaychain]
 default_command = "polkadot"
-chain_spec_path = "./scripts/zombienet/relay-chain-specs/paseo-local.plain.json"
+chain = "rococo-local"
```
JuaniRios, line 7: why use rococo instead of paseo if Polimec expects paseo?

lrazovic: Because even with the latest Paseo plain chain_spec (src: https://github.com/paseo-network/runtimes/blob/main/chain-specs/paseo-local.plain.json) - it doesn't work.

image

JuaniRios (approved):

PR #401, opened 2024-09-18
🤝  One Token Model - Pallet Funding Integration

JuaniRios:
## What?
Integrate the OTM flow in pallet-funding

## How?
- Tightly couple pallet-proxy-bonding with pallet-funding
- Abstract multiplier parameter in extrinsic to an enum called `ParticipationMode` with variants `Classic(multiplier)` and `OTM`
- Abstract the plmc bonding to a new function that takes the participation mode, and either does the classic bonding, or calls the proxy-bonding logic
- On settlement start, set the release type to `Refunded` or `Locked` in proxy-bonding pallet.
- On participation settlements, if the project failed, or a bid has some refund, call the `refund_fee` of proxy-bonding

## Testing?
In the next PR


JuaniRios:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/Polimec/polimec-node/401?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>
> <a href="https://graphite.dev/docs/merge-pull-requests">Learn more</a>

* **#401** <a href="https://app.graphite.dev/github/pr/Polimec/polimec-node/401?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#399** <a href="https://app.graphite.dev/github/pr/Polimec/polimec-node/399?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

lrazovic (approved):

pallets/funding/src/types.rs
```diff
@@ -800,12 +800,31 @@ pub mod inner {
 		pub project_id: ProjectId,
 		pub migration_origins: MigrationOrigins,
 	}
+
+	#[derive(
+		Clone, Copy, Encode, Decode, Eq, PartialEq, RuntimeDebug, TypeInfo, MaxEncodedLen, Serialize, Deserialize,
+	)]
+	pub enum ParticipationMode {
```
vstam1, line 807: I might prefer instead of passing this to the participation extrinsics, to split each participation type into two extrinsics, where the first keeps the multiplier parameter and the otm doesn't. Then you can pass a hardcoded 5 multiplier to the inner function (or pass this ParticipationMode multiplier to the inner function).

runtimes/polimec/src/lib.rs
```diff
@@ -1078,6 +1079,31 @@ impl pallet_funding::Config for Runtime {
 	type WeightInfo = weights::pallet_funding::WeightInfo<Runtime>;
 }
 
+use polimec_common::{PLMC_DECIMALS, PLMC_FOREIGN_ID, USD_DECIMALS};
+parameter_types! {
+	// Fee is defined as 1.5% of the usd_amount. Since fee is applied to the plmc amount, and that is always 5 times
+	// less than the usd_amount (multiplier of 5), we multiply the 1.5 by 5 to get 7.5%
+	pub FeePercentage: Perbill = Perbill::from_rational(75u32, 1000u32);
+	// TODO: add a real account here
+	pub FeeRecipient: AccountId = [0u8; 32].into();
```
vstam1, line 1088: Don't forget to update this account.

pallets/funding/src/functions/6_settlement.rs
```diff
@@ -146,32 +167,31 @@ impl<T: Config> Pallet<T> {
...
+				project_id,
+				HoldReason::Participation.into(),
+				&bid.bidder,
+				refunded_plmc,
+				bid.funding_asset.id(),
+			)?;
+		} else {
+			Self::release_participation_bond_for(&bid.bidder, refunded_plmc)?;
+		}
 		Self::release_funding_asset(project_id, &bid.bidder, refunded_funding_asset_amount, bid.funding_asset)?;
```
vstam1, line 186: Might be good to first do the release_funding_asset just in case the refund_fee < then min deposit amount. See previous pr for similar comment

pallets/funding/src/lib.rs
```diff
@@ -865,7 +871,7 @@ pub mod pallet {
 			jwt: UntrustedToken,
```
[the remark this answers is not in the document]
JuaniRios: - Benchmarking both cases from one extrinsic is easy
- Adding a conditional weight on top of the extrinsics based on the OTM/Classic enum is also easy
- Splitting the extrinsic into 2 after all this time is hard

I prefer to keep it as is.

pallets/funding/src/types.rs
JuaniRios: As mentioned on the other thread, I see your reasoning, but I don't think it adds enough value to justify the effort it would take in refactoring everything. If its just an extrinsic that calls the inner with the same enum, then all we're saving is the conditional weight macro on top of the extrinsic, instead of hardcoding one per extrinsic.

PR #420, opened 2024-10-14
🚬 Ethereum compatibility

JuaniRios:
## What?
- Add support for Ethereum based projects by letting users define the ethereum account that receives the tokens at the moment of participation.

## How?
- Participation info structs now have a receiving_account field which is an XCM `Junction`. This has to be either `AccountKey20` or `AccountId32`
- Issuers now define at project creation if they expect one or the other account `Junction`.
- Existing participation extrinsics now call the inner participation functions with AccountId32 == caller
- New participation extrinsics now accept a junction, and a signature. The signature will be checked against the account defined in the signature to prove ownership.

## Testing?
In the next PR: https://github.com/Polimec/polimec-node/pull/425

## Anything Else?
- For now we only support `sr25519` signatures for Polkadot/Substrate-based accounts, and `secp256k1` for Ethereum.
- The Ethereum signatures need to follow the [`eth_signTypedData_v4`](https://docs.metamask.io/wallet/reference/json-rpc-methods/eth_signtypeddata_v4/) standard: https://eips.ethereum.org/EIPS/eip-712


JuaniRios:
* **#420** <a href="https://app.graphite.dev/github/pr/Polimec/polimec-node/420?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

vstam1 (approved):
Lgtm. Cannot verify signature verification, but logic looks solid

Lederstrumpf (approved):
### Act I.
> 🚬 [...] #420

Clown

### Act II.

2024-11-07T11:11:40,474664775+01:00

Superclown
### Act III.
> ### Testing?
> In the next PR

Gigaclown :D

---

LGTM :+1: - follows EIP-191 well. Just note my comments re. recid value in sigs. I'm aware `ethers` still uses "legacy" by default - it's a minor adjustment.

pallets/funding/src/functions/misc.rs
```diff
@@ -413,6 +421,74 @@ impl<T: Config> Pallet<T> {
...
+				let message_length = message_bytes.len().to_string().into_bytes();
+				let message_prefix = b"\x19Ethereum Signed Message:\n".to_vec();
+				let full_message = [&message_prefix[..], &message_length[..], &message_bytes[..]].concat();
+				let hashed_message = keccak_256(full_message.as_slice());
+
+				match signature_bytes[64] {
+					27 => signature_bytes[64] = 0x00,
+					28 => signature_bytes[64] = 0x01,
+					_v => return Err(Error::<T>::BadReceiverAccountSignature.into()),
+				}
```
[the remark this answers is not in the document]
JuaniRios: Tried signing with Ethereum wallets using the newest signing standard EIP-712 (TypedDatav4) and it still uses the old recid

pallets/funding/src/functions/misc.rs
```diff
@@ -413,9 +421,168 @@ impl<T: Config> Pallet<T> {
...
 		<PriceProviderOf<T>>::get_decimals_aware_price(funding_asset_id, USD_DECIMALS, funding_asset_decimals)
 	}
 }
+
+pub mod typed_data_v4 {
+	use super::*;
+
+	/// Returns the first part needed for a typed data v4 message. It specifies the entity that requires the signature.
+	pub fn get_domain_separator(name: &str, version: &str, chain_id: u32, verifying_contract: &str) -> [u8; 32] {
+		let encoded_domain_type =
```
lrazovic, line 528: The Hash of ""EIP712Domain(string name,string version,uint256 chainId,address verifyingContract)" is stable, we can pre-compute it instead of computing it every time. It should look something like this

```rust
/// EIP-712 domain separator calculation
/// RFC: https://eips.ethereum.org/EIPS/eip-712
pub const DOMAIN_TYPE_HASH: &[u8; 32] = &[
    // keccak256("EIP712Domain(string name,string version,uint256 chainId,address verifyingContract)")
    0x8b, 0x73, 0xc3, 0xc6, 0x9b, 0xb8, 0xfe, 0x3d, 
    0x51, 0x2e, 0xcc, 0x4c, 0xf7, 0x59, 0xcc, 0x79, 
    0x23, 0x9f, 0x7b, 0x17, 0x9b, 0x0f, 0xfa, 0xdb, 
    0xa7, 0x2d, 0xef, 0x6c, 0x7c, 0x5d, 0x5c, 0x4c
];
```

pallets/funding/src/functions/misc.rs
```diff
@@ -413,9 +421,168 @@ impl<T: Config> Pallet<T> {
...
+		// TODO: Handle panics from copy_from_slice
+		// u32 should be converted to u256 in big endian notation
+		let chain_id_bytes: [u8; 4] = chain_id.to_be_bytes();
+		let mut encoded_chain_id: [u8; 32] = [0u8; 32];
+		encoded_chain_id[28..].copy_from_slice(&chain_id_bytes);
+		// Should be 32 bytes to comply with ABI encoding, so we pad with zeroes to the left
+		let mut encoded_contract: [u8; 32] = [0u8; 32];
+		encoded_contract[12..].copy_from_slice(hex::decode(verifying_contract).unwrap().as_slice());
+
+		let mut data = Vec::new();
```
lrazovic, line 542: Since the size of the final array is known, I would replace the `Vec` with fixed-size array. In this way we should also get rid of the `.unwrap()`. Also the `copy_from_slice` should not fails since the data is composed by components with a precise size of 32 bytes (keccak output). So the total size is fixed at 32 * 5 = 160 bytes.

A proposed solution could be:

```rust
    let mut data = [0u8; 32 * 5]; // Space for all 5 32-byte components
    
    // Copy pre-computed domain type hash
    data[..32].copy_from_slice(DOMAIN_TYPE_HASH);
    
    // Hash and copy name
    let name_hash = keccak_256(name.as_bytes());
    data[32..64].copy_from_slice(&name_hash);
    
    // Hash and copy version
    let version_hash = keccak_256(version.as_bytes());
    data[64..96].copy_from_slice(&version_hash);
    
    // Convert chain_id to big-endian bytes and pad left with zeros
    let chain_id_bytes = chain_id.to_be_bytes();
    data[124..128].copy_from_slice(&chain_id_bytes);
    
    // Decode and validate contract address
    let contract_bytes = hex::decode(verifying_contract)
        .map_err(|_| ERROR)?;
    
    if contract_bytes.len() != ADDRESS_LENGTH {
        return Err(ERROR);
    }
    
    // Copy contract address with proper padding
    data[140..160].copy_from_slice(&contract_bytes);
    
    // Calculate final hash
    keccak_256(&data)
```

pallets/funding/src/functions/misc.rs
```diff
@@ -413,9 +421,168 @@ impl<T: Config> Pallet<T> {
...
+		data.extend_from_slice(&encoded_name);
+		data.extend_from_slice(&encoded_version);
+		data.extend_from_slice(&encoded_chain_id);
+		data.extend_from_slice(&encoded_contract);
+
+		keccak_256(&data)
+	}
+
+	/// Returns the second part needed for a typed data v4 message. It specifies the message details with type information.
+	pub fn get_message(polimec_account: &str, project_id: u32, nonce: u32) -> [u8; 32] {
```
lrazovic, line 553: Similar to before, we can precompute the hash and pre-allocate a fixed size array for all components

pallets/funding/src/functions/misc.rs
```diff
@@ -413,9 +421,168 @@ impl<T: Config> Pallet<T> {
...
+		data.extend_from_slice(&encoded_message_type);
+		data.extend_from_slice(&encoded_polimec_account_string);
+		data.extend_from_slice(&encoded_project_id);
+		data.extend_from_slice(&encoded_nonce);
+
+		keccak_256(&data)
+	}
+
+	/// Returns the final message hash that will be signed by the user.
+	pub fn get_eip_712_message(polimec_account: &str, project_id: u32, nonce: u32) -> [u8; 32] {
```
lrazovic, line 577: Also here we can get rid of the `Vec`. The prefix is constant 

```
const EIP712_PREFIX: [u8; 2] = [0x19, 0x01];
const TOTAL_MESSAGE_LENGTH: usize = 2 + 32 + 32; // prefix + domain_separator + message
```

And, applying the same logic saw before

```
    // Pre-allocate fixed size array for all components
    let mut data = [0u8; TOTAL_MESSAGE_LENGTH];
    
    // Copy the EIP-712 prefix
    data[..2].copy_from_slice(&EIP712_PREFIX);
    
    // Get domain separator and copy it
    let domain_separator = get_domain_separator(
        "Polimec",
        "1",
        1,
        "0000000000000000000000000000000000003344"
    )?;
    data[2..34].copy_from_slice(&domain_separator);
    
    // Get message hash and copy it
    let message = get_message(polimec_account, project_id, nonce)?;
    data[34..66].copy_from_slice(&message);
    
    // Calculate final hash
   keccak_256(&data)
```

lrazovic (changes requested):
The logic looks good! I proposed few improvements on the EIP-712 side, so we can get rid of the `Vec`. Also, we should add benchmarks for the new receiving account extrinsics since we're doing extra computation for the signature verification (ECDSA/sr25519) in there.

lrazovic (approved):

JuaniRios:
### Merge activity

* **Jan 16, 7:09 AM EST**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/Polimec/polimec-node/420).

PR #431, opened 2024-12-04
🎗️ Feeless calls

JuaniRios:
## What?
Refund the extrinsic execution fee for calls that require a JWT

## Why?
The fee payment is used to avoid spamming and stalling the chain, but all JWT calls from regular KYCs need to either pay a fee with OTM, or lock up PLMC.

And the other ones are locked behind an institutional JWT, so even if they act malicious we could check who they are and take action.

This should provide better UX, since you can say you spend 100USDT and spend 100USDT, instead of 100.03

## How?
Instead of returning `Ok(())` return 
```
Ok(PostDispatchInfo{
   actual_weight: None,
   pays_fee: Pays::No,
})
```

## Testing?
Tests don't include fee payments for extrinsics. So only way I can see to test this is creating a network with chopsticks/zombienet and trying it out ourselves.

For now it hasn't been tested, but its a simple change so should work.

### DIFF
--- a/pallets/funding/src/functions/1_application.rs
+++ b/pallets/funding/src/functions/1_application.rs
@@ -1,6 +1,7 @@
 #![allow(clippy::wildcard_imports)]
 #![allow(clippy::type_complexity)]
 
+use sp_runtime::offchain::http::Method::Post;
 use super::*;
 
 impl<T: Config> Pallet<T> {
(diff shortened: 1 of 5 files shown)


lrazovic (approved):

JuaniRios:
### Merge activity

* **Jan 16, 8:51 AM EST**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/Polimec/polimec-node/431).

PR #433, opened 2024-12-11
Fix fungibles API

JuaniRios:
## What?
- The fungibles API should return the universally recognized `Location` of an asset, and not where its stored locally.
For example, if a parachain holds DOT in their assets pallet, the fungibles API should return the id of DOT as `(parents: 1, interior: Here)`, and not the internal location of the assets pallet.

- For this we should change the id used for foreign assets to be this universal id, and not a u32.

## Why?
- Wallets will start using this API, and it needs to behave as expected.
- We want to use this API when writing  multichain chopsticks e2e tests.

## Testing?
In the intergration tests


runtimes/polimec/src/lib.rs
```diff
@@ -1325,29 +1315,30 @@ mod benches {
...
-						.filter(|(_, balance)| balance > &0)
-				)?,
-				// collect pallet_assets (ForeignAssets)
-				convert::<_, _, _, _, xcm_config::ForeignAssetsConvertedConcreteId>(
-					ForeignAssets::account_balances(account)
-						.iter()
-						.filter(|(_, balance)| balance > &0)
-				)?,
-			].concat().into())
+			// Ok([
```
lrazovic, line 1318: Restored in PR #434

lrazovic (approved):

lrazovic (approved):
LGTM - I leave a deeper XCM config review to @vstam1

JuaniRios:
### Merge activity

* **Jan 16, 7:09 AM EST**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/Polimec/polimec-node/433).

PR #462, opened 2025-02-26
Upgrade to stable2412-2

dastansam:


dastansam:
will remove identity pallet and add a migration now

Issue #469, opened 2025-03-14
Investigate why `BlockNumberProvider::set_block_number` doesn't work


dastansam:
when working on #468, I stumbled on an issue where `<T as Config>::BlockNumberProvider::set_block_number()` does not set the block number in both unit and integration tests. This is why it's currently still using `System::set_block_number()` and overwriting validation data in `ParachainSystem`.

PR #471, opened 2025-03-14
✨ Prepare pallets for async backing

dastansam:
## What?

Introduces `BlockNumberProvider` type for pallets

## Why?

So that we can use a more reliable, consistent notion of block number, since with async backing system block number [becomes](https://wiki.polkadot.network/docs/maintain-guides-async-backing#async-backing-terminology-and-parameters) nondeterministic

## How?

Relay chain's block number is used where possible, i.e funding, linear-release, and we rely on system block number in other cases

## Testing?

Changes were made to `Instantiator` so that it works for both integration and unit tests, since now integration tests Runtime uses relay chain's block number instead of chain's

## Anything Else?
https://github.com/Polimec/polimec-node/issues/469

### DIFF
--- a/pallets/dispenser/src/lib.rs
+++ b/pallets/dispenser/src/lib.rs
@@ -55,7 +55,7 @@ pub mod pallet {
 		pallet_prelude::{ValueQuery, *},
 		PalletId,
 	};
-	use frame_system::pallet_prelude::*;
+	use frame_system::pallet_prelude::{BlockNumberFor, *};
 	use sp_runtime::{
 		traits::{AccountIdConversion, CheckedDiv},
 		Saturating,
--- a/pallets/funding/src/lib.rs
+++ b/pallets/funding/src/lib.rs
@@ -79,7 +79,6 @@ use frame_support::{
 	},
 	BoundedVec, PalletId, WeakBoundedVec,
 };
-use frame_system::pallet_prelude::BlockNumberFor;
 pub use pallet::*;
 use polimec_common::{
 	credentials::{Cid, Did, EnsureOriginWithCredentials, InvestorType, UntrustedToken},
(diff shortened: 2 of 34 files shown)


pallets/funding/src/lib.rs
```diff
@@ -134,9 +136,11 @@ pub type VestingOf<T> = pallet_linear_release::Pallet<T>;
 pub type BlockNumberToBalanceOf<T> = <T as pallet_linear_release::Config>::BlockNumberToBalance;
 pub type RuntimeHoldReasonOf<T> = <T as Config>::RuntimeHoldReason;
 pub type PriceProviderOf<T> = <T as Config>::PriceProvider;
+pub type BlockNumberFor<T> = <<T as Config>::BlockNumberProvider as BlockNumberProvider>::BlockNumber;
 
 #[frame_support::pallet]
 pub mod pallet {
+	use super::BlockNumberFor;
```
lrazovic, line 143: Isn't this also part of the `use super::*;` two lines below?

pallets/dispenser/src/lib.rs
```diff
@@ -115,6 +115,9 @@ pub mod pallet {
 		/// The Whitelisted policy for the dispenser. Users' credentials should have the same
 		/// policy to be eligible for token dispensing.
 		type WhitelistedPolicy: Get<Cid>;
+
+		// /// Block number provider.
+		// type BlockNumberProvider: BlockNumberProvider<BlockNumber = BlockNumberFor<Self>>;
```
lrazovic, line 120: What happened here?

lrazovic (approved):
Just 2 nits, but the approach seems straightforward!

lrazovic:
### Merge activity

* **May 2, 7:23 AM EDT**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/Polimec/polimec-node/471).

PR #495, opened 2025-08-26
Code comments

dastansam:
## What?

- more inline comments
- removed unused event/error variants
- added more comments for types


lrazovic (approved):
Final LGTM

